PR #145, opened 2021-02-28
Replaced with shape theme

### LINKED ISSUE #120: Define shape theming
## Kind (Required)

- Improvement

 ## Overview (Required)

- Right now it is defined for each widget, but it should be possible to set it with theme. 

 ## Links	

- https://github.com/DroidKaigi/conference-app-2021/blob/610ad01d00a673211e9bcef716762c367cdb53d6/uicomponent-compose/core/src/main/java/io/github/droidkaigi/feeder/core/theme/Shape.kt#L8

NUmeroAndDev:
## Issue
- close #120

## Overview (Required)
- Define shape theme and replaced using shape theme in some components.

## Links
-

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/uicomponent-compose/core/src/main/java/io/github/droidkaigi/feeder/core/theme/Shape.kt
+++ b/uicomponent-compose/core/src/main/java/io/github/droidkaigi/feeder/core/theme/Shape.kt
@@ -1,11 +1,19 @@
 package io.github.droidkaigi.feeder.core.theme
 
-import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.foundation.shape.CutCornerShape
 import androidx.compose.material.Shapes
 import androidx.compose.ui.unit.dp
 
 val shapes = Shapes(
-    small = RoundedCornerShape(4.dp),
-    medium = RoundedCornerShape(4.dp),
-    large = RoundedCornerShape(0.dp)
+    small = CutCornerShape(
+        topStart = 8.dp,
+        bottomEnd = 8.dp
+    ),
+    medium = CutCornerShape(
+        topStart = 8.dp,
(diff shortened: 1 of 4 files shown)


uicomponent-compose/core/src/main/java/io/github/droidkaigi/feeder/core/theme/Shape.kt
```diff
@@ -1,11 +1,19 @@
...
 
 val shapes = Shapes(
-    small = RoundedCornerShape(4.dp),
-    medium = RoundedCornerShape(4.dp),
-    large = RoundedCornerShape(0.dp)
+    small = CutCornerShape(
+        topStart = 8.dp,
+        bottomEnd = 8.dp
+    ),
+    medium = CutCornerShape(
```
NUmeroAndDev, line 12: There was no medium components(ex. Card, Dialog) in Figma, so temporarily made the same as small shape.

takahirom:
👀

uicomponent-compose/feed/src/main/java/io/github/droidkaigi/feeder/feed/FeedItem.kt
```diff
@@ -54,10 +53,7 @@ fun FeedItem(
                     }
                     .background(
                         color = feedItem.media.color(),
-                        shape = CutCornerShape(
-                            topStart = 8.dp,
-                            bottomEnd = 8.dp
-                        )
+                        shape = MaterialTheme.shapes.small
```
takahirom, line 56: I didn't know that shape could be used this way! Thank you!

takahirom:
LGTM! Thanks for first contribute 🥇

PR #156, opened 2021-02-28
Allow to run workflows on contributors's repos

jmatsu:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
- The apparent behavior won't be changed but the assemble workflow (it's the core of the build system) will be run on `pull_request` event. So contributors can modify the workflows if needed.
- This PR may fail due to the difference from the current default branch

## Links
-

## Screenshot
Before | After
:--: | :--:
 |


jmatsu:
Let me merge

PR #161, opened 2021-02-28
use action model pattern for navigation

### LINKED ISSUE #111: Use Action model pattern for navigation
## Kind (Required)

- Improvement

 ## Overview (Required)

- This is the implementation pattern in compose-sample.

 ## Links	

- https://github.com/android/compose-samples/blob/2cff65334ea023956a58f215a13b6a55c7e1e18f/Owl/app/src/main/java/com/example/owl/ui/NavGraph.kt#L75

ked4ma:
## Issue
- close #111

## Overview (Required)
- use action model pattern for navigation on AppComponent.kt

- Unable to pass method reference of actions directly...
  - for DrawerContent, close drawer operation is needed
  - for onDetailClick, context instance is needed

Because of above, I wrap method of actions and pass it.

## Links
-

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/uicomponent-compose/main/src/main/java/io/github/droidkaigi/feeder/AppContent.kt
+++ b/uicomponent-compose/main/src/main/java/io/github/droidkaigi/feeder/AppContent.kt
@@ -47,10 +47,9 @@ fun AppContent(
         drawerShape = MaterialTheme.shapes.large.copy(all = CornerSize(0.dp)),
         drawerContent = {
             DrawerContent { route ->
-                actions.onSelectDrawerItem(route) {
-                    coroutineScope.launch {
-                        drawerState.close()
-                    }
+                actions.onSelectDrawerItem(route)
+                coroutineScope.launch {
+                    drawerState.close()
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


takahirom:
👀

uicomponent-compose/main/src/main/java/io/github/droidkaigi/feeder/AppContent.kt
```diff
@@ -37,15 +40,14 @@ fun AppContent(
...
     ModalDrawer(
         modifier = modifier,
         drawerState = drawerState,
         drawerShape = MaterialTheme.shapes.large.copy(all = CornerSize(0.dp)),
         drawerContent = {
             DrawerContent { route ->
-                try {
-                    navController.navigate(route)
-                } finally {
+                actions.onSelectDrawerItem(route) {
```
takahirom, line 50: Thank you for leaving the original code! 
I don't think I need try catch anymore, so maybe I can write it like this

```kotlin
actions.onSelectDrawerItem(route)
coroutineScope.launch {
     drawerState.close()
}
```

takahirom (approved):
Thanks!

uicomponent-compose/main/src/main/java/io/github/droidkaigi/feeder/AppContent.kt
ked4ma: Thanks a lot for your review!

> I don't think I need try catch anymore

You'are right. It's my wrong!
I simplified this part!!
https://github.com/DroidKaigi/conference-app-2021/pull/161/commits/95eab98f8a0de26aa3aa15ef75eee991fbe31b69

takahirom:
This makes navigation easier to list! Thank you!

PR #175, opened 2021-02-28
Highlighted navigation drawer item

### LINKED ISSUE #163: Highlighted item  in Navigation Drawer according to the currently selected destination
## Kind (Required)

- Improvement

 ## Overview (Required)
Now the Home item is always highlighted in Navigation Drawer.

How about making it the item is highlighted according to the currently selected destination?
For example, this is what the DroidKaigi2020 app looks like.

https://user-images.githubusercontent.com/19250035/109412185-e3231480-79e9-11eb-9ca3-761ee52bb425.mp4

 ## Links	

- https://developer.android.com/jetpack/compose/navigation?hl=ja#bottom-nav

Dai1678:
## Issue
- close #163 

## Overview (Required)
- Highlighted drawer item when you clicked drawer item

## What I won't do in this PR
- Highlighted drawer item when you clicked **Tab item**

Sorry, I couldn't think of a better solution...

## Links
- FYI: https://developer.android.com/jetpack/compose/navigation#bottom-nav

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/uicomponent-compose/main/src/main/java/io/github/droidkaigi/feeder/AppContent.kt
+++ b/uicomponent-compose/main/src/main/java/io/github/droidkaigi/feeder/AppContent.kt
@@ -53,7 +53,7 @@ fun AppContent(
             val currentOtherTab = navBackStackEntry?.arguments?.getString("otherTab")
             val currentRoute = currentFeedTab?.let { "feed/$it" }
                 ?: currentOtherTab?.let { "other/$it" }
-                ?: "feed/${FeedTabs.Home.routePath}"
+                ?: DrawerContents.HOME.route
             DrawerContent(currentRoute) { route ->
                 actions.onSelectDrawerItem(route)
                 coroutineScope.launch {
--- a/uicomponent-compose/main/src/main/java/io/github/droidkaigi/feeder/AppContent.kt
+++ b/uicomponent-compose/main/src/main/java/io/github/droidkaigi/feeder/AppContent.kt
@@ -8,6 +8,7 @@ import androidx.compose.material.MaterialTheme
 import androidx.compose.material.ModalDrawer
 import androidx.compose.material.rememberDrawerState
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
 import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.platform.LocalContext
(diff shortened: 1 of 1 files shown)


Dai1678:
> What I won't do in this PR
Highlighted drawer item when you clicked Tab item

I'm sorry. I couldn't think of a good solution.
Does anyone have a good solution?
Or please help me as another welcome contribute issue!

takahirom:
👀

uicomponent-compose/main/src/main/java/io/github/droidkaigi/feeder/AppContent.kt
```diff
@@ -46,7 +48,13 @@ fun AppContent(
         drawerState = drawerState,
         drawerShape = MaterialTheme.shapes.large.copy(all = CornerSize(0.dp)),
         drawerContent = {
-            DrawerContent { route ->
+            val navBackStackEntry by navController.currentBackStackEntryAsState()
```
takahirom, line 51: How about this implementation? 👀 

```kotlin
 var currentDestination by remember { mutableStateOf<NavDestination>()}
    navController.addOnDestinationChangedListener { controller, destination: NavDestination, arguments ->
        currentDestination = destination
    }
```

takahirom: I will check it a little more. 👀

takahirom: If you do this implementation (addOnDestinationChangedListener), a lot of listeners will be added, so you have to remember or do something. 👀

takahirom:
Currently, I think this is the best implementation 👍 
By the way, do you think it should be reflected when tabs are selected?

Dai1678:
I think what complicates this issue is that the Drawer and Tab have the same role.
However, I think it would be difficult to change the design drastically, so maybe we don't need to do the highlighting item in Drawer.

Quit using `DrawerButton.isSelected`

I'm interested in hearing other people's opinions.

takahirom:
Well, it may be better to create an issue. Can you make it if you like?

Dai1678:
Okay. I will created a new issue.
Thank you so much!

Dai1678:
@takahirom 
I created a new issue #180 

Will this PR be merged?

takahirom:
OK! Thank you for thinking about this!

takahirom (approved):

PR #188, opened 2021-03-01
stop generating BuildConfig

### LINKED ISSUE #125: Disable BuildConfig generate for compile time
## Kind (Required)

- Improvement

 ## Overview (Required)

- If there is BuildConfig, the Java compilation will run, so I want to eliminate BuildConfig in the module.

 ## Links	

-

tkmusic1976:
## Issue
- close #125

## Overview (Required)
- Set `buildConfig` property to false at android.buildFeatures of android/build.gradle.

## Links
You can get a detailed explanation of this property from the link below.
https://developer.android.com/reference/tools/gradle-api/7.0/com/android/build/api/dsl/BuildFeatures#buildconfig

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -72,6 +72,7 @@ android {
     }
     buildFeatures {
         compose true
+        buildConfig false
     }
     composeOptions {
         kotlinCompilerExtensionVersion "${compose_version}"


takahirom (approved):
LGTM! I didn't know that we can disable BuildConfig very easily!

PR #189, opened 2021-03-01
Prevent backdrop from moving up and down on other screen

### LINKED ISSUE #85: Prevent backdrop from moving up and down on other screen
## Kind (Required)

- Improvement

 ## Overview (Required)

https://user-images.githubusercontent.com/1386930/108645294-bce70b80-74f5-11eb-81ea-ff6683fb600a.mp4

- On other screens, I don't want to move the backdrop because there is no content even if I lower the backdrop down. Now it goes up and down 1dp.

 ## Links	

- https://github.com/DroidKaigi/conference-app-2021/blob/main/uicomponent-compose/other/src/main/java/io/github/droidkaigi/feeder/other/OtherScreen.kt#L89

ue-moo:
## Issue
- close #85 

## Overview (Required)
-  Fix `BackdropScaffold#gesturesEnabled` value to `false`  on other screen.

## Links
-

## Screenshot
https://user-images.githubusercontent.com/70118729/109441257-22984200-7a78-11eb-9763-603cdc004faf.mp4

### DIFF
--- a/uicomponent-compose/other/src/main/java/io/github/droidkaigi/feeder/other/OtherScreen.kt
+++ b/uicomponent-compose/other/src/main/java/io/github/droidkaigi/feeder/other/OtherScreen.kt
@@ -83,6 +83,7 @@ fun OtherScreen(
     Column {
         val density = LocalDensity.current
         BackdropScaffold(
+            gesturesEnabled = false,
             backLayerBackgroundColor = MaterialTheme.colors.primary,
             scaffoldState = scaffoldState,
             backLayerContent = {


uicomponent-compose/other/src/main/java/io/github/droidkaigi/feeder/other/OtherScreen.kt
```diff
@@ -83,6 +83,7 @@ fun OtherScreen(
     Column {
         val density = LocalDensity.current
         BackdropScaffold(
+            gesturesEnabled = false,
```
takahirom, line 86: I didn't notice this propertyy! Thank you!

takahirom (approved):

PR #195, opened 2021-03-01
Add English support for MultiLangText #174

### LINKED ISSUE #174: Handling about MultiLangText model
## Kind (Required)

- Improvement

 ## Overview (Required)

We've introduced MultiLangText to handle multi-language.
But we don't support displaying English text yet.

https://github.com/DroidKaigi/conference-app-2021/blob/main/uicomponent-compose/feed/src/main/java/io/github/droidkaigi/feeder/feed/FeedItem.kt#L91

 ## Links	

We can reference a implementation of the previous version of the conference app.
https://github.com/DroidKaigi/conference-app-2020/blob/master/model/src/jvmMain/kotlin/Lang.kt

https://github.com/DroidKaigi/conference-app-2020/blob/2c703ff896b9b58f07300d1c55b2d622c19203a1/model/src/commonMain/kotlin/io/github/droidkaigi/confsched2020/model/LocaledString.kt

-

yetanothercheer:
## Issue
- close #174 

## Overview (Required)
- Implement MultiLangText by Locale
- Change default jaTitle to currentLangTitle

## Links
-

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/model/src/commonMain/kotlin/io/github/droidkaigi/feeder/Lang.kt
+++ b/model/src/commonMain/kotlin/io/github/droidkaigi/feeder/Lang.kt
@@ -1,0 +1,10 @@
+package io.github.droidkaigi.feeder
+
+import java.util.Locale
+
+enum class Lang {
+    JA,
+    EN;
+}
+
+fun defaultLang() = if (Locale.getDefault() == Locale.JAPAN) Lang.JA else Lang.EN
--- a/model/src/commonMain/kotlin/io/github/droidkaigi/feeder/MultiLangText.kt
+++ b/model/src/commonMain/kotlin/io/github/droidkaigi/feeder/MultiLangText.kt
@@ -3,4 +3,14 @@ package io.github.droidkaigi.feeder
 class MultiLangText(
     val jaTitle: String,
     val enTitle: String,
-)
+) {
+    val currentLangTitle get() = getByLang(defaultLang())
+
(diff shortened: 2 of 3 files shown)


mhidaka:
👀

mhidaka (approved):
That‘s fantastic! LGTM 💯

model/src/commonMain/kotlin/io/github/droidkaigi/feeder/Lang.kt
```diff
@@ -0,0 +1,10 @@
+package io.github.droidkaigi.feeder
+
+import java.util.Locale
```
RyuNen344, line 3: @yetanothercheer  @mhidaka 
sorry for comment to closed PR 🙇 
in commonMain module should't use `java.util`, right? (because, ios moudle won't be able to use this)

mhidaka: 👀 @RyuNen344  if you consider sharing a module with iOS, It is need some kind of another ways.
we think this approach is fine for the PR. Thank you for your great opinion 👍 

Can you make a new issue ✍ if you like ?

RyuNen344: @mhidaka 
thank you for reply! I will create a issue!

PR #196, opened 2021-03-01
Keep selected tab through rotation at FeedScreen

### LINKED ISSUE #127: Keep selected tab through screen rotation
## Kind (Required)

- Improvement

 ## Overview (Required)

- Since this is the state of FeedScreen, it may be possible to keep it in ViewModel

 ## Links	

-

Index197511:
## Issue
- close #127 

## Overview (Required)
- Tab selection returns to Home every time when the screen is rotated, so it is now keeped in the ViewModel.

## Links
-

## Screenshot
Before | After
:--: | :--:
 |


Index197511:
When an item is selected from the Drawer, FeedScreen receives the selected tab as InitialSelectedTab, but how do I propagate this as the initial state to the ViewModel?

Index197511:
Currently, I have implemented it only in FeedScreen, but should I implement it in OtherScreen as well?

mhidaka:
👀

mhidaka:
It's good question💡
@Index197511 Don't worries, It is walcome thing that you'll just implement the FeedScreen to improve keeping selected tab within rotation☎

By the way, we're glad that the number of contributors will increase if you create a new issue for other screens to fix the same problem.

Index197511:
Thank you!
I've finished implementation and make an issue to implement the selected tabs from the Drawer and something similar in OtherScreen.

Index197511:
@mhidaka 
Thank you!
I've finished implementation and make an issue to implement the selected tabs from the Drawer and something similar in OtherScreen.

mhidaka:
Great! Thank you for your additional work. and The review of this PR is waiting for A reviewer in charge to get freetime. please wait a minite 😉

takahirom:
👀

takahirom:
> When an item is selected from the Drawer, FeedScreen receives the selected tab as InitialSelectedTab, but how do I propagate this as the initial state to the ViewModel? @takahirom

How about this way?

```kotlin
    effectFlow.collectInLaunchedEffect { effect ->
        when (effect) {
            is FeedViewModel.Effect.ErrorMessage -> {
                scaffoldState.snackbarHostState.showSnackbar(
                    effect.appError.getReadableMessage(context)
                )
            }
        }
    }
    LaunchedEffect(initialSelectedTab) {
        dispatch(...)
    }
```

Index197511:
It looks very good.
However, with this way, won't the LanchedEffect be called every time the screen is rotated and the Tab will be reset?

Index197511:
It looks very good.
However, with this way, won't the LanchedEffect be called every time the screen is rotated and the Tab will be reset?
@takahirom

takahirom:
I will check more 👀

takahirom:
It was my mistake. Perhaps the selected tabs are unmanageable in this scope. 👀 
I'm very sorry, is it possible to move the tab management process to AppContent and use savedInstanceState()?

Index197511:
It may take a while, but I'll try it.

Index197511:
Would it be better to create another issue and implement it there? @takahirom

takahirom:
I have assigned you to this issue. Therefore, you can add an implementation to this PR or issue another PR. 👍 
https://github.com/DroidKaigi/conference-app-2021/issues/198

Index197511:
I'm thinking if I try to save FeedTabs within onSavedInstanceState, Activity must own selected tab in the FeedScreen. I'm not sure how to do this, as having a ViewModel in the Activity seems subtle.
@takahirom

takahirom:
Can you use navigation? (I tried. but this is not fixed format commit)
https://github.com/DroidKaigi/conference-app-2021/commit/7a907ac9602267443265a912744d7c69ef121e48

Index197511:
Thank you for your help!
I moved the SelectedTab that I had in ViewModel to Navigation.
It's working well.

Index197511:
Thank you for your help! @takahirom 
I moved the SelectedTab that I had in ViewModel to Navigation.
It's working well.
I've finished implementation for this PR.

takahirom (approved):
Thank you for tackling 👍 LGTM

PR #204, opened 2021-03-01
Use org.jetbrains.kotlinx:kotlinx-coroutines-core from Dep.kt

Pluu:
## Issue
- close none

## Overview (Required)
- Use org.jetbrains.kotlinx:kotlinx-coroutines-core from Dep.kt

## Links
-

## Screenshot
Before | After
:--: | :--:
 |


takahirom (approved):
LGTM! Do you have eagle eyes?🦅

PR #221, opened 2021-03-01
Use Arbor for logging library

### LINKED ISSUE #63: Use Arbor for logging library
## Kind (Required)

- Improvement

 ## Overview (Required)

- In this project, println is used as a log in the context of Kotlin Multiplatform.
- It seems that this library can be used on Kotlin Multiplatform like Timber, so I would like to use it.

 ## Links	

- https://github.com/ToxicBakery/Arbor

kmachida12345:
## Issue
- close #63

## Overview (Required)
- use Arbor.d instead of println
- use Startup to initialize Arbor

## Links
-

## Screenshot
Before | After
:--: | :--:
 |


tomoya0x00:
👀

android/src/main/java/io/github/droidkaigi/feeder/Initializers.kt
```diff
@@ -15,3 +17,13 @@ class FirebaseInitializer : Initializer<Firebase> {
         return emptyList()
     }
 }
+
+class ArborInitializer : Initializer<Unit> {
+    override fun create(context: Context) {
+        Arbor.sow(Seedling())
```
tomoya0x00, line 23: It looks work fine 👍 
But, How do you think about initialize with `LogCatSeedling` instead of `Seedling` ? 👀 

Please refer below:
https://github.com/ToxicBakery/Arbor#api

image

tomoya0x00 (commented):
@kmachida12345 Could you check my comment and fix the conflict? 🙇 Thanks!

kmachida12345:
thank you for your review and comment🙇‍♂️ I'll fix them soon!

takahirom:
👀

android/src/main/java/io/github/droidkaigi/feeder/ArborInitializer.kt
```diff
@@ -0,0 +1,16 @@
+package io.github.droidkaigi.feeder
+
+import android.content.Context
+import androidx.startup.Initializer
+import com.toxicbakery.logging.Arbor
+import com.toxicbakery.logging.LogCatSeedling
+
+class ArborInitializer : Initializer<Unit> {
```
takahirom, line 8: Initializer！✨

takahirom (approved):
Thank you for Arbor.sow() 🌱 LGTM!

PR #235, opened 2021-03-02
Modify scale type of feed image in the FeedItem from center crop to center inside

### LINKED ISSUE #88: Prototype image scale type
## Kind (Required)

- Improvement

 ## Overview (Required)

- Designers seem to be worried about image proportions.
I think the feed image of this app is center crop now, but I would like to try patterns such as center inside, take screenshots, ask the designer for a good pattern, and implement it.

 ## Links	

-
image

maho-ya:
## Issue
- close #88

## Overview (Required)
- This is the feed screen image after changing the scale type from center crop to center inside. Please consider which is better.

## Links
-

## Screenshot
Before (center crop) | After (center inside)
:--: | :--:
 | 
 |


takahirom:
👀

maho-ya:
@takahirom 
Hi, I would like to talk about the discussion that took place below.
https://github.com/DroidKaigi/conference-app-2021/pull/243#discussion_r585580544

In order to get closer to Figma's design, I have the following image in mind.
I also rounded the image.

Figma | Suggestion
:--: | :--:
 | 

I set both the height and width to 96dp, as shown below. (The background is added for clarity as an example.)

size

What do you think?

takahirom:
Thank you! I'll just talk to the designer today based on this 🙏

takahirom:
Excuse me, can you take a screenshot of the center crop version as well?

maho-ya:
Below is a comparison image of center crop and center inside.

center crop | center inside
:--: | :--:
 |

takahirom:
Thank you!!  🙏

takahirom:
Thank you! I talked to the designer. 👍 
Since the size of the image looks different, it was said that the center crop is good.
If you like, I'd like to include just the corner change, but what do you think?

maho-ya:
I think center crop is better because the image size is not uniform.
I just added rounded corners to image.

uicomponent-compose/feed/src/main/java/io/github/droidkaigi/feeder/feed/FeedItem.kt
```diff
@@ -80,6 +82,7 @@ fun FeedItem(
                     bottom.linkTo(parent.bottom, 12.dp)
                 }
                 .width(96.dp)
+                .clip(shape = RoundedCornerShape(4.dp))
 //                .aspectRatio(16F / 9F)
```
maho-ya, line 86: Should I delete this comment?
It does not seem to represent the current situation.

maho-ya:
https://github.com/DroidKaigi/conference-app-2021/blob/7b30053b542c7772afd1a76983b70f9710bf1447/uicomponent-compose/core/src/main/java/io/github/droidkaigi/feeder/core/NetworkImage.kt#L26

Do you know if this Box is working?

uicomponent-compose/feed/src/main/java/io/github/droidkaigi/feeder/feed/FeedItem.kt
takahirom: Yes can you delete it?

```suggestion
```

takahirom: Anyway, I wii merge this PR!

takahirom (approved):
Thank you for your cooperation!! LGTM 👍

maho-ya:
Thank you!

PR #257, opened 2021-03-03
Remove "java.util.Locale" usage from model module commonMain

### LINKED ISSUE #200: getDefaultLocale by platform implementation
## Kind (Required)
- Other

 ## Overview (Required)

- Pure kotlin code is preferable for common modules.
- https://github.com/DroidKaigi/conference-app-2021/pull/195/files/0c217968280001a98c0664d445654608e6f421b6#r584475947
- I think it needs expect/actual fun

 ## Links	

- [kotlin natvie mapping](https://kotlinlang.org/docs/native-objc-interop.html#mappings)

Gazyu:
## Issue
- close #200 

## Overview (Required)
- Remove "java.util.Locale" usage from model module commonMain.

### DIFF
--- a/model/src/main/java/io/github/droidkaigi/feeder/Locale.kt
+++ b/model/src/main/java/io/github/droidkaigi/feeder/Locale.kt
@@ -6,4 +6,3 @@ actual fun getDefaultLocale(): Locale =
     } else {
         Locale.OTHER
     }
-
--- a/model/src/main/java/io/github/droidkaigi/feeder/Locale.kt
+++ b/model/src/main/java/io/github/droidkaigi/feeder/Locale.kt
@@ -1,0 +1,9 @@
+package io.github.droidkaigi.feeder
+
+actual fun getDefaultLocale(): Locale =
+    if (java.util.Locale.getDefault() == java.util.Locale.JAPAN) {
+        Locale.JAPAN
+    } else {
+        Locale.OTHER
+    }
+
--- a/model/src/commonMain/kotlin/io/github/droidkaigi/feeder/Lang.kt
+++ b/model/src/commonMain/kotlin/io/github/droidkaigi/feeder/Lang.kt
@@ -1,10 +1,8 @@
 package io.github.droidkaigi.feeder
 
-import java.util.Locale
-
 enum class Lang {
     JA,
     EN;
 }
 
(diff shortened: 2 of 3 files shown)


tomoya0x00:
👀

model/src/main/java/io/github/droidkaigi/feeder/Locale.kt
```diff
@@ -0,0 +1,9 @@
+package io.github.droidkaigi.feeder
+
+actual fun getDefaultLocale(): Locale =
+    if (java.util.Locale.getDefault() == java.util.Locale.JAPAN) {
+        Locale.JAPAN
+    } else {
+        Locale.OTHER
+    }
+
```
tomoya0x00, line 9: Could you remove this empty line? 🙏 

image

model/src/main/java/io/github/droidkaigi/feeder/Locale.kt
```diff
@@ -0,0 +1,9 @@
+package io.github.droidkaigi.feeder
+
+actual fun getDefaultLocale(): Locale =
+    if (java.util.Locale.getDefault() == java.util.Locale.JAPAN) {
```
tomoya0x00, line 4: How do you think about using `import java.util.Locale` ? 👀

tomoya0x00 (commented):
@Gazyu Could you check my comments? Thanks 🙏

model/src/main/java/io/github/droidkaigi/feeder/Locale.kt
Gazyu: @tomoya0x00 
If I use `import java.util.Locale`, it caused an error because the "Locale Class" is the same name.
So I used full path.
If there is a better way, please could you let me know.

tomoya0x00: @Gazyu
Oh, I see. You are right!
Thank you for your explanation 🙇 

We don't need to change this point 👍

model/src/main/java/io/github/droidkaigi/feeder/Locale.kt
Gazyu: @tomoya0x00 
I fix it!

tomoya0x00 (approved):
LGTM! Thank you for your contribution 🎉

PR #251, opened 2021-03-02
Transition animation when switching tab content

fornewid:
## Issue
- close #79

## Overview (Required)
- Unfortunately, `navigation-compose` does not supports transition animation between `composable`s.
- To achieve transition animation between tab contents, 
   we should NOT navigate new `composable` when tab is selected.
   Instead, we need to update `selectedTab` state inside the current FeedScreen on each `onSelectedTab()` events.
- And material motion was not implemented in material compose.
   So I use `Crossfade` instead.
   (If okay, let's implement [FadeThrough](https://material.io/design/motion/the-motion-system.html#fade-through) animation as the next task.)

## Links
- ~~https://issuetracker.google.com/issues/172112072~~
   
- https://developer.android.com/jetpack/compose/animation#crossfade

## Screenshot
Before | After
:--: | :--:
 |

Issue #87, opened 2021-02-23
Animate the frame to move when a tab is selected


takahirom:
## Kind (Required)

- Improvement

 ## Overview (Required)

- Perhaps this is very difficult, but can you animate this corner cut frame to move?

![image](https://user-images.githubusercontent.com/1386930/108819312-0c186380-75fe-11eb-976d-b97f2f00b232.png)

 ## Links	

-

ntsk:
:raising_hand_man:

I don't know if it can be done, but I will try it with reference to [JetSnack](https://github.com/android/compose-samples/tree/main/Jetsnack).

<img src=https://user-images.githubusercontent.com/5681281/109499756-a97d0700-7ad8-11eb-9dbe-c82979bcff95.gif width=300>

takahirom:
🆒 
Please try it 🙏 If it's difficult, you can always give up.

PR #261, opened 2021-03-03
Transition animation when switching tab content

### LINKED ISSUE #79: Transition animation when switching tab content
## Kind (Required)

- Improvement

 ## Overview (Required)
- According to Material Design's motion system, this app should use Fade through in such cases.

 ## Links	

- https://material.io/design/motion/the-motion-system.html#fade-through

fornewid:
## Issue
- close #79

## Overview (Required)
- Unfortunately, `navigation-compose` does not supports transition animation between `composable`s.
- To achieve transition animation between tab contents, 
   we should NOT navigate new `composable` when tab is selected.
   Instead, we need to update `selectedTab` state inside the current `FeedScreen`/`OtherScreen` on each events.
- material motion was not implemented in `material-compose`.
   So use `Crossfade` instead.
- :new: Add `DrawerContentState`, `rememberRoutePath()` to fix issues commented on https://github.com/DroidKaigi/conference-app-2021/pull/251#issuecomment-789305708.

## Links
- ~~https://issuetracker.google.com/issues/172112072~~
   
- https://developer.android.com/jetpack/compose/animation#crossfade

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/uicomponent-compose/main/src/main/java/io/github/droidkaigi/feeder/AppContent.kt
+++ b/uicomponent-compose/main/src/main/java/io/github/droidkaigi/feeder/AppContent.kt
@@ -51,7 +51,7 @@ fun AppContent(
         "https://" + LocalContext.current.getString(R.string.deep_link_host) +
             LocalContext.current.getString(R.string.deep_link_path)
     val actions = remember(navController, drawerContentState) {
-        AppActions(navController, drawerContentState)
+        AppActions(navController)
     }
     ModalDrawer(
         modifier = modifier,
@@ -59,7 +59,9 @@ fun AppContent(
         drawerShape = MaterialTheme.shapes.large.copy(all = CornerSize(0.dp)),
         drawerContent = {
             DrawerContent(drawerContentState.currentValue) { route ->
-                actions.onSelectDrawerItem(route)
+                if (drawerContentState.selectDrawerContent(route)) {
+                    actions.onSelectDrawerItem(route)
+                }
                 coroutineScope.launch {
                     drawerState.close()
                 }
@@ -82,14 +84,14 @@ fun AppContent(
                         ?: FeedTabs.Home.routePath
                 )
                 val selectedTab = FeedTabs.ofRoutePath(routePath.value)
-                actions.onSelectTab(selectedTab)
+                drawerContentState.onSelectDrawerContent(selectedTab)
                 val context = LocalContext.current
                 FeedScreen(
                     onNavigationIconClick = onNavigationIconClick,
                     selectedTab = selectedTab,
                     onSelectedTab = { feedTabs ->
                         routePath.value = feedTabs.routePath
-                        actions.onSelectTab(feedTabs)
+                        drawerContentState.onSelectDrawerContent(feedTabs)
                     },
                     onDetailClick = { feedItem: FeedItem ->
                         actions.onSelectFeed(context, feedItem)
@@ -110,12 +112,12 @@ fun AppContent(
                         ?: OtherTabs.AboutThisApp.routePath
                 )
                 val selectedTab = OtherTabs.ofRoutePath(routePath.value)
-                actions.onSelectOtherTab(selectedTab)
+                drawerContentState.onSelectDrawerContent(selectedTab)
                 OtherScreen(
                     selectedTab = selectedTab,
                     onSelectTab = { otherTabs ->
                         routePath.value = otherTabs.routePath
-                        actions.onSelectOtherTab(otherTabs)
+                        drawerContentState.onSelectDrawerContent(otherTabs)
                     },
                     onNavigationIconClick = onNavigationIconClick
                 )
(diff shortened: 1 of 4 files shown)


takahirom:
I have a question! Since this is an animation implementation of the contents of the tab, will it remain even if the Compose side support is released?
https://android-review.googlesource.com/c/platform/frameworks/support/+/1508708

takahirom:
If it remains, AppAction and Tab actions have different scopes, so it may be better not to combine them. 
Anyway, thanks for the good animations and solutions! ✨

fornewid:
@takahirom
First, thanks for sharing. :smile:

Looking at the code in the link, I think... 🤔 
If we **only** need animation between tab content, we don't need to merge this.

But,
If AppBar also needs animation, it would be better to merge it. (https://github.com/DroidKaigi/conference-app-2021/issues/87)

Please tell me your opinions. :bow:

takahirom:
Thank you for your feedback.  I'll take a look tomorrow morning.🙏

fornewid:
I applies the comments above.
`DrawerContentState` is separated from `AppActions`

And also set different `navOptions` depending on `DrawerContents`.
Now navigation behavior is same as [droidkaigi 2020](https://github.com/DroidKaigi/conference-app-2020).

uicomponent-compose/other/src/main/java/io/github/droidkaigi/feeder/other/OtherScreen.kt
```diff
@@ -176,6 +153,39 @@ private fun AppBar(
     }
 }
 
+@Composable
+private fun Content(
```
takahirom, line 157: I think it depends on your taste, but it seems that it will take time to know which part is content, so how about BackdropFrontLayerContent?

takahirom: You can merge without changing. 👍

uicomponent-compose/main/src/main/java/io/github/droidkaigi/feeder/AppContent.kt
```diff
@@ -49,20 +50,18 @@ fun AppContent(
     val deepLinkUri =
         "https://" + LocalContext.current.getString(R.string.deep_link_host) +
             LocalContext.current.getString(R.string.deep_link_path)
-    val actions = remember(navController) { AppActions(navController) }
+    val actions = remember(navController, drawerContentState) {
```
takahirom, line 53: I have a question! Do you need this key? 🙋

fornewid: Sorry for missing it. I'll fix it now. :smile:

uicomponent-compose/other/src/main/java/io/github/droidkaigi/feeder/other/OtherScreen.kt
fornewid: Good idea! I'll change name.

fornewid:
Thanks for comments! I applied code review.

uicomponent-compose/main/src/main/java/io/github/droidkaigi/feeder/AppContent.kt
```diff
@@ -114,8 +127,19 @@ fun AppContent(
 }
 
 private class AppActions(navController: NavHostController) {
-    val onSelectDrawerItem: (String) -> Unit = { route ->
-        navController.navigate(route)
+    val onSelectDrawerItem: (DrawerContents) -> Unit = { contents ->
+        navController.navigate(contents.route) {
+            popUpTo(navController.graph.startDestination) {
```
takahirom, line 132: Maybe it's a difficult code for those who are new to it, so could you please add a comment? 🙇

takahirom:
I'm sorry, I'm not sure, but why can I fade while transitioning by navigation when selecting a tab? 🙋

takahirom:
I'm sorry it took me a while to understand. 🙇

fornewid:
When selecting a tab, only tab state changes without navigation.
if tab state changes, transition animation is activated by the crossfade. :bow:

uicomponent-compose/main/src/main/java/io/github/droidkaigi/feeder/AppContent.kt
fornewid: I added a comment. Would it be okay? :eyes:

uicomponent-compose/main/src/main/java/io/github/droidkaigi/feeder/AppContent.kt
```diff
@@ -80,13 +79,20 @@ fun AppContent(
...
+                drawerContentState.onSelectDrawerContent(selectedTab)
                 val context = LocalContext.current
                 FeedScreen(
                     onNavigationIconClick = onNavigationIconClick,
-                    selectedTab = FeedTabs.ofRoutePath(feedType),
-                    onSelectedTab = { feedTabs -> actions.onSelectTab(feedTabs) },
+                    selectedTab = selectedTab,
+                    onSelectedTab = { feedTabs ->
+                        routePath.value = feedTabs.routePath
+                        drawerContentState.onSelectDrawerContent(feedTabs)
```
takahirom, line 94: Thank you! How about putting a comment like this? 🙏 

```suggestion
                        // We don't use navigation transitions here for animation.
                        routePath.value = feedTabs.routePath
                        drawerContentState.onSelectDrawerContent(feedTabs)
```

fornewid: Thanks for suggestion. I added a comment!

takahirom (approved):
Thanks for exploring the structure of this app and library and making changes! 👍

PR #263, opened 2021-03-03
Setup Flipper

### LINKED ISSUE #91: Try Flipper
## Kind (Required)

- Improvement

 ## Overview (Required)

- I've never used it, but I heard it's a good library.
- Probably should we use OkHttp engline?
- https://github.com/DroidKaigi/conference-app-2021/blob/be4293122b884513b9aa02f088965a85832742e5/data/api/src/main/java/io/github/droidkaigi/feeder/data/ApiModule.kt#L25

 ## Links	
- https://fbflipper.com/

kafumi:
## Issue

- close #91

## Overview (Required)

- Integrated [Flipper](https://fbflipper.com/) for debugging app
- Set up following plugins
  - [Network](https://fbflipper.com/docs/setup/network-plugin)
  - [SharedPreferences](https://fbflipper.com/docs/setup/shared-preferences-plugin)
  - [Databases](https://fbflipper.com/docs/setup/databases-plugin)

## Implementation notes

- Set up Flipper 0.76.0 although its latest is 0.78.0. App crashed when I tried Flipper 0.77.0 and 0.78.0 because of https://github.com/facebook/flipper/issues/1968. We should keep using 0.76.0 until the issue is resolved
- Added Flipper dependencies only in `debug` source set without using its "no-op" package to follow [Flipper's recommendation](https://github.com/facebook/flipper/issues/1968)
  > Please note that our `flipper-noop` package provides a limited subset of the APIs provided by the `flipper` package and does not provide any plugin stubs. It is recommended that you keep all Flipper instantiation code in a separate build variant to ensure it doesn't accidentally make it into your production builds.
- OkHttp was implicitly used as the engine of Ktor (selected automatically by `ServiceLoader`) even before this PR. I explicitly specify it when configuring `HttpClient` to set `FlipperOkhttpInterceptor` to `OkHttpClient`
- I added `@Singleton` annotation to the functions of `ApiModule` to make sure that `HttpClient` is created only once. According to the 2nd example of [component scopes](https://developer.android.com/training/dependency-injection/hilt-android#component-scopes) document, it seem necessary to add `@Singleton` annotation to the functions. Actually I found `HttpClient` was instantiated twice without this annotation

## Screenshot

https://user-images.githubusercontent.com/12084705/109826908-9efc7200-7c7e-11eb-8eb4-4ee5aef0cf50.mov

### DIFF
--- a/buildSrc/src/main/java/Dep.kt
+++ b/buildSrc/src/main/java/Dep.kt
@@ -116,6 +116,7 @@ object Dep {
         //      We can upgrade it to newer version once following issue is resolved.
         //      https://github.com/facebook/flipper/issues/1968
         const val flipper = "com.facebook.flipper:flipper:0.76.0"
+        const val networkPlugin = "com.facebook.flipper:flipper-network-plugin:0.76.0"
         const val soLoader = "com.facebook.soloader:soloader:0.10.1"
     }
 }
--- a/buildSrc/src/main/java/Dep.kt
+++ b/buildSrc/src/main/java/Dep.kt
@@ -110,4 +110,12 @@ object Dep {
         const val jvm = "com.ToxicBakery.logging:arbor-jvm:1.35.72"
         const val android = "com.ToxicBakery.logging:arbor-android:1.35.72"
     }
+
+    object Flipper {
+        // TODO We use Flipper 0.76.0 for now because app crashes with 0.77.0 and 0.78.0.
+        //      We can upgrade it to newer version once following issue is resolved.
+        //      https://github.com/facebook/flipper/issues/1968
+        const val flipper = "com.facebook.flipper:flipper:0.76.0"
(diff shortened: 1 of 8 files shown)


tomoya0x00:
👀

buildSrc/src/main/java/Dep.kt
```diff
@@ -110,4 +110,13 @@ object Dep {
         const val jvm = "com.ToxicBakery.logging:arbor-jvm:1.35.72"
         const val android = "com.ToxicBakery.logging:arbor-android:1.35.72"
     }
+
+    object Flipper {
+        // TODO We use Flipper 0.76.0 for now because app crashes with 0.77.0 and 0.78.0.
```
tomoya0x00, line 115: This comment is so helpful for others 👍

tomoya0x00 (approved):
Thanks for your contribution, especially kindly implementation notes 📝 
It was so helpful for our review 👍

PR #288, opened 2021-03-06
Implement getDefaultLocale for iOS

### LINKED ISSUE #272: implement Kotlin Native version of getDefaultLocale()
## Kind (Required)

- Improvement

 ## Overview (Required)

 - Please somehow use Kotlin Native to determine the locale for Japan or not. Perhaps the iOS development documentation will be helpful.
 - Add one actual function to iosMain and you're done.

You can check it by this.

./gradlew :model:compileKotlinIos

e: /Users/takahirom/git/connference-app-2021-news/model/src/commonMain/kotlin/io/github/droidkaigi/feeder/Locale.kt: (8, 12): Expected function 'getDefaultLocale' has no actual declaration in module <conference-app-2021:model> for Native

image

 ## Links	

-

myihsan:
## Issue
- close #272 

## Overview (Required)
- Determine `Locale` by current `countryCode`.

## Links
- https://developer.apple.com/documentation/foundation/nslocale/1409990-currentlocale?language=objc
- https://developer.apple.com/documentation/foundation/nslocale/1643060-countrycode?language=objc

### DIFF
--- a/model/src/iosMain/kotlin/io/github/droidkaigi/feeder/Locale.kt
+++ b/model/src/iosMain/kotlin/io/github/droidkaigi/feeder/Locale.kt
@@ -1,0 +1,12 @@
+package io.github.droidkaigi.feeder
+
+import platform.Foundation.NSLocale
+import platform.Foundation.countryCode
+import platform.Foundation.currentLocale
+
+actual fun getDefaultLocale(): Locale =
+    if (NSLocale.currentLocale.countryCode == "JP") {
+        Locale.JAPAN
+    } else {
+        Locale.OTHER
+    }


takahirom:
👀

takahirom (approved):
I didn't know iOS API! Thank you!

model/src/iosMain/kotlin/io/github/droidkaigi/feeder/Locale.kt
```diff
@@ -0,0 +1,12 @@
+package io.github.droidkaigi.feeder
+
+import platform.Foundation.NSLocale
+import platform.Foundation.countryCode
+import platform.Foundation.currentLocale
+
+actual fun getDefaultLocale(): Locale =
+    if (NSLocale.currentLocale.countryCode == "JP") {
```
takahirom, line 8: There is no need to fix it, but if you want to start a new line here, it may be better to use block according to the style guide.
https://developer.android.com/kotlin/style-guide#expression_functions

myihsan: Thanks for your advice.
I thought it is normal according to the implementation for Android, but it seems like that this was only because the code didn't fit on a single line.

PR #297, opened 2021-03-06
Animate the frame to move when a tab is selected

ntsk:
## Issue
- close https://github.com/DroidKaigi/conference-app-2021/issues/87

## Overview (Required)
- Added animation when selecting tabs.
- ScrollableTabRow did not seem to implement an option like [TabLayout::INDICATOR_GRAVITY_STRETCH](https://developer.android.com/reference/com/google/android/material/tabs/TabLayout#INDICATOR_GRAVITY_STRETCH).
- The ScrollableTabRow indicator is drawn over the text, so I created a custom ScrollableTabRow that's indicator can be displayed behind the text.

## Links
- #87

## Screenshot
Before | After
:--: | :--:
 |


uicomponent-compose/core/src/main/java/io/github/droidkaigi/feeder/core/ScrollableTabRow.kt
```diff
@@ -0,0 +1,379 @@
+package io.github.droidkaigi.feeder.core
+
+/**
+ * Copied from [androidx.compose.material.ScrollableTabRow] to support background indicator.
+ */
```
ntsk, line 5: This change may make it difficult to keep up with library updates. 😔 
Is there a good way to customize the Composable function?

takahirom:
👀

uicomponent-compose/core/src/main/java/io/github/droidkaigi/feeder/core/ScrollableTabRow.kt
```diff
@@ -0,0 +1,379 @@
...
+                tabPlaceables.fastForEach {
+                    it.placeRelative(left, 0)
+                    left += it.width
+                }
+
+                subcompose(TabSlots.ForegroundIndicator) {
+                    foregroundIndicator(tabPositions)
+                }.fastForEach {
+                    it.measure(Constraints.fixed(layoutWidth, layoutHeight)).placeRelative(0, 0)
+                }
```
ntsk, line 173: In the original ScrollTabRow, the indicator is drawn after the text, so it will be displayed above.
This custom ScrollTabRow allows to display the Indicator behind the text.
https://github.com/androidx/androidx/blob/master/compose/material/material/src/commonMain/kotlin/androidx/compose/material/TabRow.kt#L258-L291

ntsk (commented):
I've tried to implement it, but what do you think? 
If you have a better idea than this, I will try it!

takahirom:
Thank you for the wonderful animation. It may take some time for me to understand. 👀

uicomponent-compose/core/src/main/java/io/github/droidkaigi/feeder/core/ScrollableTabRow.kt
takahirom: 👍

takahirom: Super cool!! I'm learning a lot. 🆒

uicomponent-compose/core/src/main/java/io/github/droidkaigi/feeder/core/ScrollableTabRow.kt
takahirom: I learned a lot from this PR. Thank you very much.
How about proxying with SubcomposeLayout?

```kotlin
fun ScrollableTabRowWithForeground() {
       val indicatorTabPosition by remember{ mutableStateOf() }
       val tabRowPlaceables = subcompose(SlotsEnum.Main) {
            ScrollableTabRow(
                ...
              indicator: @Composable (tabPositions: List<TabPosition>) -> Unit = @Composable { tabPositions ->
                  val indicatorPosition = Modifier.tabIndicatorOffset(tabPositions[selectedTabIndex])
                  TabRowDefaults.Indicator(
                       Modifier.tabIndicatorOffset(indicatorPosition)
                   )
                   indicatorTabPosition = indicatorPosition
              },
            )
        }.map {
            it.measure(constraints)
        }
        layout(...) {
            // layout indicator with indicatorPosition and tabRowPlaceables
        }
```
}

ntsk: Thanks for the suggestion! I'll try it ! 😃

takahirom:
This PR animation is great, so why not merge it once and proceed without copying the code? 👀

ntsk:
@takahirom 
Thank you very much!
I'm trying the suggestions, but it may take some time to complete, so it would be nice to be able to merge once. :pray: 
I'll continue to try if I can do it without copying the code.

takahirom:
Thank you!!!

PR #301, opened 2021-03-06
Use Const Value for Route String

### LINKED ISSUE #229: Use const value for route string
## Kind (Required)

- Improvement

 ## Overview (Required)

- We are using `feed/` `other/`
- Can you use top-level const value for this in AppContent?
- https://github.com/DroidKaigi/conference-app-2021/blob/main/uicomponent-compose/main/src/main/java/io/github/droidkaigi/feeder/AppContent.kt#L72

 ## Links	

-

k-tomoyasu:
## Issue
- close #229

## Overview (Required)
- use const value for route path string `feed/` , `other/`

## Links
-

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/uicomponent-compose/main/src/main/java/io/github/droidkaigi/feeder/AppContent.kt
+++ b/uicomponent-compose/main/src/main/java/io/github/droidkaigi/feeder/AppContent.kt
@@ -32,6 +32,9 @@ import io.github.droidkaigi.feeder.other.OtherScreen
 import io.github.droidkaigi.feeder.other.OtherTabs
 import kotlinx.coroutines.launch
 
+private const val FEED_PATH = "feed/"
+private const val OTHER_PATH = "other/"
+
 @Composable
 fun AppContent(
     modifier: Modifier = Modifier,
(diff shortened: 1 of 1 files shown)


uicomponent-compose/main/src/main/java/io/github/droidkaigi/feeder/AppContent.kt
```diff
@@ -98,8 +101,8 @@ fun AppContent(
                 )
             }
             composable(
-                route = "other/{otherTab}",
-                deepLinks = listOf(navDeepLink { uriPattern = "$deepLinkUri/other/{otherTab}" }),
+                route = "$OTHER_PATH{otherTab}",
+                deepLinks = listOf(navDeepLink { uriPattern = "$deepLinkUri/$OTHER_PATH{otherTab}" }),
```
takahirom, line 105: Thank you for cleaning!
Almost LGTM 👍 
Can you add a new line here?
Our CI style check tool fails. 👀

k-tomoyasu: overlooked linter advice. thanks!

takahirom (approved):
LGTM 👍

PR #346, opened 2021-03-11
Separate Ktor settings from platform-specific and common

### LINKED ISSUE #94: Separate Ktor settings from platform-specific and common
## Kind (Required)

- Improvement

 ## Overview (Required)

- Right now I'm making it below, but this is code that is only used on Android and should be shared on Multiplatform.

https://github.com/DroidKaigi/conference-app-2021/blob/be4293122b884513b9aa02f088965a85832742e5/data/api/src/main/java/io/github/droidkaigi/feeder/data/ApiModule.kt#L23-L51

- For example, if you can pass platform-specific settings as shown below, you can share logic with Multiplatform.

```kotlin
class MyClient<T : HttpClientEngineConfig>(
    engine: HttpClientEngineFactory<T>,
    configBlock: T.()
    -> Unit,
) {
    private val headerValue = "fuga"
    private val httpClient = HttpClient(engine) {
        defaultRequest {
            header("hoge", headerValue)
        }
        engine(configBlock)
    }
}
```

 ## Links	

-

watanavex:
## Issue
- close #94

## Overview (Required)
- Add ktors `HttpClient` wrapper
- Perform common configuration via this wrapper

... and Add iOS target and dependencies

## Links
-

## Screenshot
🙅‍♂️

### DIFF
--- a/data/api/build.gradle
+++ b/data/api/build.gradle
@@ -19,20 +19,8 @@ tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile.class).all {
 
 kotlin {
     android()
-//    jvm()
-    // currently firebase-auth not supported iOS
-    // https://github.com/GitLiveApp/firebase-kotlin-sdk/issues/133
-//    if (System.getProperty("idea.active") == "true") {
-//        // Fix IntelliJ Unresolved reference
-//        def onPhone = System.getenv("SDK_NAME")?.startsWith("iphoneos") ?: false
-//        if (onPhone) {
-//            iosArm64("ios")
-//        } else {
-//            iosX64("ios")
-//        }
-//    } else {
-//        ios()
-//    }
+    ios()
+
     sourceSets {
         commonMain.dependencies {
             api project(":model")
@@ -87,6 +75,9 @@ kotlin {
             kotlin {
                 srcDir 'src/androidTest'
             }
+        }
+        iosMain.dependencies {
+            implementation Dep.Ktor.ios
         }
 //            // For instrumentation tests
 //            androidTestImplementation  "com.google.dagger:hilt-android-testing:2.31.2-alpha"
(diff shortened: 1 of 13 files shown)


takahirom:
👀

data/api/build.gradle
```diff
@@ -88,6 +76,9 @@ kotlin {
                 srcDir 'src/androidTest'
             }
         }
+        iosMain.dependencies {
+            implementation Dep.Ktor.ios
```
takahirom, line 80: 🆒

data/api/src/commonMain/kotlin/io/github/droidkaigi/feeder/data/NetworkService.kt
```diff
@@ -0,0 +1,57 @@
...
+import io.ktor.client.features.json.serializer.KotlinxSerializer
+import io.ktor.client.features.logging.LogLevel
+import io.ktor.client.features.logging.Logger
+import io.ktor.client.features.logging.Logging
+import io.ktor.client.request.headers
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.modules.SerializersModule
+import kotlinx.serialization.modules.contextual
+
+class NetworkService private constructor(val httpClient: HttpClient) {
```
takahirom, line 18: Thank you for realizing the commonality!
I thought that this class would access the property of httpClient every time, and it seems that it is not good for encapsulation, so I thought that it is better to distribute HttpClient at present. 💭 
If you plan to make HttpClient a private property and encapsulate it in NetworkService, this is fine, but it seems that we will need to create comments and issues. 🙇

takahirom: However, it can be fixed immediately, and iOS development is prioritized over encapsulation issues, so we will merge this PR. 👍

takahirom (approved):
Thank you for advancing the development of KMM!!!

data/api/src/commonMain/kotlin/io/github/droidkaigi/feeder/data/NetworkService.kt
watanavex: @takahirom  I thought it was a problem too.
I would like to improve it at another time.

Thank you for always reviewing!

PR #348, opened 2021-03-11
Bundle DroidKaigi FM in app

### LINKED ISSUE #118: Bundle DroidKaigi FM in app
## Kind (Required)

- Improvement

 ## Overview (Required)
 - Can we play it in the app?
 - The mp3 url seems to have regularity and may be usable.
 - https://droidkaigi.jp/fm/audio/droidkaigi-fm_6.mp3

 ## Links	

-

akihiro-n:
## Issue
- close #118 

## Overview (Required)
- add simple audio play button
- generate an mp3 URL
- implemented DroidKaigi FM player

### DIFF
--- a/model/build.gradle
+++ b/model/build.gradle
@@ -15,6 +15,8 @@ kotlin {
         android
         commonMain.dependencies {
             implementation Dep.Kotlin.bom
+            implementation Dep.Ktor.bom
+            implementation Dep.Ktor.serialization
             api project.dependencies.platform(Dep.Coroutines.bom)
             api Dep.Coroutines.core
             api Dep.datetime
--- a/uicomponent-compose/main/src/main/java/io/github/droidkaigi/feeder/viewmodel/RealFeedViewModel.kt
+++ b/uicomponent-compose/main/src/main/java/io/github/droidkaigi/feeder/viewmodel/RealFeedViewModel.kt
@@ -16,9 +16,9 @@ import io.github.droidkaigi.feeder.getContents
 import io.github.droidkaigi.feeder.orEmptyContents
 import io.github.droidkaigi.feeder.repository.FeedRepository
 import io.github.droidkaigi.feeder.toLoadState
-import kotlinx.coroutines.Dispatchers
 import javax.annotation.meta.Exhaustive
 import javax.inject.Inject
+import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableStateFlow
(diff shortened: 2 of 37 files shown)


model/build.gradle
```diff
@@ -15,6 +15,8 @@ kotlin {
         android
         commonMain.dependencies {
             implementation Dep.Kotlin.bom
+            implementation Dep.Ktor.bom
+            implementation Dep.Ktor.serialization
```
akihiro-n, line 19: The reason I added this is because it uses the Url class !

takahirom:
👀

model/build.gradle
takahirom: If you add it to the constructor of the podcast class, you may not need to depend on it in the model module. 👀

uicomponent-compose/main/src/main/java/io/github/droidkaigi/feeder/viewmodel/RealFeedViewModel.kt
```diff
@@ -40,6 +46,16 @@ class RealFeedViewModel @Inject constructor(
...
 
+    private val mediaPlayer by lazy {
+        MediaPlayer().apply {
+            setAudioAttributes(
+                AudioAttributes.Builder()
+                    .setLegacyStreamType(AudioManager.STREAM_MUSIC)
+                    .build()
+            )
+        }
+    }
```
takahirom, line 57: Thanks for implementing it to make it playable! 👍 
If you put the processing related to Player in FeedViewModel, it seems that it will be too large, so it seems better to separate it into another ViewModel. There is no need to respond with this pull request.

takahirom:
It seems that CI is failing because the format is broken. 🙇 
Can you run spotlessKotlinApply task from here? 🙏 
image

model/build.gradle
akihiro-n: Thanks for the review!
That's a good way. Thanks for the idea!

uicomponent-compose/main/src/main/java/io/github/droidkaigi/feeder/viewmodel/RealFeedViewModel.kt
akihiro-n: I understand. Thanks for your advice!

takahirom:
Thank you for format. 🙇 Can you check this? 🙏 
e: /home/runner/work/conference-app-2021/conference-app-2021/uicomponent-compose/feed/src/androidTest/java/io/github/droidkaigi/feeder/feed/FeedItemTest.kt: (109, 17): No value passed for parameter 'onClickPlayPodcastButton'

akihiro-n:
OK! I will confirm.

tomoya0x00:
👀

takahirom:
👀

takahirom (approved):
Thank you for implementing a new experience to DroidKaigi App!

PR #351, opened 2021-03-12
Get app version name on About this app screen

### LINKED ISSUE #312: Get app version name on About this app screen
## Kind (Required)
- Improvement

 ## Overview (Required)
- Related issue #308 
- Get app version name on About this app screen

t-nonomura:
## Issue
- close #312

## Overview (Required)
- I got the version name of the app and displayed it on the About this app screen.
- I have created AppVersions.kt that describes the version information of the app.
- ~The debug build adds the "-debug" suffix.~
- ~Since the release build cannot be done, the display in the release build cannot be confirmed.:worried:~
## Links
-

## Screenshot
Before(fixed string "1.0.0") | After
:--: | :--:
 |

### DIFF
--- a/uicomponent-compose/other/build.gradle
+++ b/uicomponent-compose/other/build.gradle
@@ -11,13 +11,8 @@ plugins {
 apply from: rootProject.file("gradle/android.gradle")
 
 android {
-    buildTypes {
-        debug {
-            resValue "string", "app_version_name", "$AppVersions.versionName-debug"
-        }
-        release {
-            resValue "string", "app_version_name", AppVersions.versionName
-        }
+    defaultConfig {
+        resValue "string", "app_version_name", AppVersions.versionName
     }
     compileOptions {
         sourceCompatibility JavaVersion.VERSION_1_8
--- a/uicomponent-compose/other/build.gradle
+++ b/uicomponent-compose/other/build.gradle
@@ -1,3 +1,4 @@
+import io.github.droidkaigi.feeder.AppVersions
 import io.github.droidkaigi.feeder.Dep
 import io.github.droidkaigi.feeder.Versions
 // TODO: make this library Kotlin MPP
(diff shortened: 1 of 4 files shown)


tomoya0x00:
👀

uicomponent-compose/other/build.gradle
```diff
@@ -10,6 +11,14 @@ plugins {
 apply from: rootProject.file("gradle/android.gradle")
 
 android {
+    buildTypes {
+        debug {
+            resValue "string", "app_version_name", "$AppVersions.versionName-debug"
+        }
+        release {
```
tomoya0x00, line 18: Thank you for tackling the issue that how to refer to the application version name from the library module 👍 

In my opinion, I think we can just define `resValue "string", "app_version_name", AppVersions.versionName` at `defaultConfig`. I think it is enough for this case.

Because `debug-buildType` at library module doesn't always mean `debug-buildType` at app module.
If app module defined other buildTypes and fall back to `debug-buildType` as library modules,  the problem will occur.

How do you think about it? 👀

tomoya0x00 (commented):
@t-nonomura Almost LGTM 🎖️ Could you check my comment? Thanks 👍

uicomponent-compose/other/build.gradle
t-nonomura: > Because debug-buildType at library module doesn't always mean debug-buildType at app module.
If app module defined other buildTypes and fall back to debug-buildType on library modules, the problem will occur.

That's true.
Thank you for your advice, looks good.
I'll fix it!

t-nonomura: I have fixed it, so please review it again:pray:

tomoya0x00 (approved):
Thank you for your quick fix and contribution ❤️  LGTM!

PR #366, opened 2021-03-14
Feature/re catch error 364

### LINKED ISSUE #364: Show error snackbar when refresh error
## Kind (Required)

- Improvement

 ## Overview (Required)

 - `try-catch` the following and send it to the `effectChannel`.
https://github.com/DroidKaigi/conference-app-2021/blob/2aca45d4be764b67dd9b56057b9fededc3939532/uicomponent-compose/main/src/main/java/io/github/droidkaigi/feeder/viewmodel/RealFeedViewModel.kt#L45

```kotlin
effectChannel.send(FeedViewModel.Effect.ErrorMessage(error.e))
```

 - This can be checked using airplane mode.

 ## Links	

-

ked4ma:
## Issue
- close #364 

## Overview (Required)
- show error snackbar when refresh feeds.
- show snackbar again and again, if errors are occurred when click reload button on it.

## Links
-

### DIFF
--- a/uicomponent-compose/main/src/main/java/io/github/droidkaigi/feeder/viewmodel/RealFeedViewModel.kt
+++ b/uicomponent-compose/main/src/main/java/io/github/droidkaigi/feeder/viewmodel/RealFeedViewModel.kt
@@ -17,6 +17,8 @@ import io.github.droidkaigi.feeder.getContents
 import io.github.droidkaigi.feeder.orEmptyContents
 import io.github.droidkaigi.feeder.repository.FeedRepository
 import io.github.droidkaigi.feeder.toLoadState
+import javax.annotation.meta.Exhaustive
+import javax.inject.Inject
 import kotlinx.coroutines.Dispatchers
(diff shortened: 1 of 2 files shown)


uicomponent-compose/main/src/main/java/io/github/droidkaigi/feeder/viewmodel/RealFeedViewModel.kt
```diff
@@ -148,9 +149,17 @@ class RealFeedViewModel @Inject constructor(
...
                 }
                 is FeedViewModel.Event.ReloadContent -> {
-                    repository.refresh()
+                    refreshFeed()
                 }
             }
         }
     }
+
+    private suspend fun refreshFeed() {
```
takahirom, line 158: Thank you for sharing it with the function!👍  With this name, it looks like an update of the display, so the name refleshRepository may not be misunderstood.  There is no need to fix it in this PR.

takahirom (approved):
LGTM!

PR #375, opened 2021-03-14
Separate MediaPlayer from RealFeedViewModel

### LINKED ISSUE #357: Separate MediaPlayer from RealFeedViewModel
## Kind (Required)

- Improvement

 ## Overview (Required)

- Putting it in makes the FeedViewModel bigger, and its functionality is completely platform-dependent, so I'd like to separate it.
 - How about FmPlayerViewModel etc.?

 ## Links	

- https://github.com/DroidKaigi/conference-app-2021/pull/348/files#diff-869051b887b09c40b445ee5ea2a14f99e5035876760b5468626d45138366a25eR49

akihiro-n:
## Issue
- close #357 

## Overview (Required)
- Add FmPlayerViewModel.
- Remove Media Player from RealFeedViewModel.

### DIFF
--- a/uicomponent-compose/main/src/main/java/io/github/droidkaigi/feeder/viewmodel/RealFeedViewModel.kt
+++ b/uicomponent-compose/main/src/main/java/io/github/droidkaigi/feeder/viewmodel/RealFeedViewModel.kt
@@ -1,8 +1,5 @@
 package io.github.droidkaigi.feeder.viewmodel
 
-import android.media.AudioAttributes
-import android.media.AudioManager
-import android.media.MediaPlayer
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import dagger.hilt.android.lifecycle.HiltViewModel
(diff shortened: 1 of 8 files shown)


takahirom:
👀

uicomponent-compose/main/src/main/java/io/github/droidkaigi/feeder/viewmodel/RealFeedViewModel.kt
```diff
@@ -135,18 +102,13 @@ class RealFeedViewModel @Inject constructor(
...
                         url = event.feedItem.podcastLink(),
                         isPlaying = isPlaying.not()
                     )
-                    if (newState.isPlaying) {
-                        playPodcast(newState.url, state?.url == newState.url)
-                    } else {
-                        pausePodcast()
-                    }
                     playingPodcastState.value = newState
+                    effectChannel.send(FeedViewModel.Effect.ControlFmPlayer(newState))
```
takahirom, line 111: Thank you for separating!  👍 
For example, if you want to play in the background, it will be difficult if the feedViewModel manages the playback state. 👀 
Therefore, is it possible to move the playback status as well?

akihiro-n: indeed! That state management method is a better way 👍 
I will fix it now. Thank you for the advice!

takahirom (approved):
LGTM ⏯️ Thank you for refactoring!

PR #410, opened 2021-04-09
Add statusBarsPadding on AppContent

### LINKED ISSUE #230: Switch status bar transparent when navigation drawer open
## Kind (Required)

- Improvement

 ## Overview (Required)

- Currently, we use always semi-transparent background.
- But navigation drawer of material component android switch transparent and semi-transparent when open drawer.
- 
image

 ## Links	

-

yasuhiroki:
## Issue

- close #230

## Overview (Required)

- Add `statusBarsPadding` on AppContent to avoid overlying status bar on app

If my solution about "switch status bar transparent when navigation drawer open" is not correct, please let me know.

### Light

View | Before | After
:--: | :--: | :--:
Top |  | 
Drawer |  | 

### Dark

View | Before | After
:--: | :--: | :--:
Top |  | 
Drawer |  |

### DIFF
--- a/uicomponent-compose/feed/src/main/java/io/github/droidkaigi/feeder/feed/FeedScreen.kt
+++ b/uicomponent-compose/feed/src/main/java/io/github/droidkaigi/feeder/feed/FeedScreen.kt
@@ -43,7 +43,6 @@ import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.painter.Painter
 import androidx.compose.ui.platform.LocalContext
-import androidx.compose.ui.platform.LocalDensity
 import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.semantics.semantics
 import androidx.compose.ui.tooling.preview.Preview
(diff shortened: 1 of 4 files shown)


takahirom:
Thank you for approaching!
What I wanted to do was make the StatusBar translucent when the NavigationDrawer was open, and always make the StatusBar transparent.
For example, it has the following form.

```
if (navigationDrawer.isOpened) {
   statusBar.color = semiTransparent
} else {
   statusBar.color = transparent
}
```

In this PR, the Status Bar is semi-transparent, but it seems that the content is not displayed under the Status Bar. Also, there seems to be no code to switch as above.
I think this implementation is so difficult that I don't even know how to do it, so if you want to give up, unassign it, so please tell me 🙇

yasuhiroki:
@takahirom
Thank you for your response.
Let me check because I still don't understand exactly.
Do you want it to look like the following images?

Drawer Open (A)| Drawer Open (B) |
 :---: | :---: |
 | 

About "Drawer Open (A)", in my understanding, most apps have a UI like this.
"Drawer Open (B)" is a screenshot that is built using`<item name="android:statusBarColor">@android:color/transparent</item>`.
"Drawer Open (B)" status bar hide some icon. So, I feel it is better "Drawer Open (A)"  than "Drawer Open (B)".

takahirom:
I imagine this kind of behavior. It seems to be different from both A and B. 👀 (from 2019 app)

When not drawer opened.
image

When drawer opened.
image

yasuhiroki:
@takahirom 
Thank you for your checking!
I understand clearly. I'll try implementing it asap by new pull request.

PR #411, opened 2021-04-14
Make it possible to reference the types of imported modules in iOS

watanavex:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
- now, IOS code cannot resolve the type of imported modules.
- For example, `db` module cannot resolve the type of the `model` module.
- It happens only in the IDE display. We can build from the command line.

I restored the flag in `gradle.properties`.

I've had success with sync in my environment.
I don't fully understand these issues.
However, I think that maybe the flag in `gradle.properties` is necessary.

## Links
- https://kotlinlang.org/docs/migrating-multiplatform-project-to-14.html#try-the-hierarchical-project-structure

## Screenshot
🙅‍♂️

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -21,6 +21,5 @@ android.enableJetifier=true
 kotlin.code.style=official
 # hierarchy structure support
 # https://kotlinlang.org/docs/mpp-share-on-platforms.html#share-code-on-similar-platforms
-# This will cause sync issue. Probably libraries not supported
-# kotlin.mpp.enableGranularSourceSetsMetadata=true
-# kotlin.native.enableDependencyPropagation=false
+kotlin.mpp.enableGranularSourceSetsMetadata=true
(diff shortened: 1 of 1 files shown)


watanavex:
最新のKMMプラグインで新規プロジェクトの動作を確認したところ、デフォルトではこれらのフラグ付与されなくなりました。
```
kotlin.mpp.enableGranularSourceSetsMetadata=true
kotlin.native.enableDependencyPropagation=false
```
しかし、共通モジュールを新しく追加すると、自動でこれらのフラグが付与されます。。。

takahirom:
Thank you for reporting! I will check in my Android Studio 👀

takahirom:
I'm not sure why I had a problem before and disabled it, but it doesn't seem to be a problem right now. 👀 
Thank you!

PR #396, opened 2021-03-20
DI framework for ios

watanavex:
## Issue
- #353

## Overview (Required)
- I adopted Koin
- Android targets keeps Dagger Hilt
- This PR, DB and API modules are supported
- Repository module has another issue (#395), I'll support with it later

## Links
- https://insert-koin.io

## Screenshot
🙅‍♂️

PR #415, opened 2021-04-17
[proposal] Resolve Firebase auth in iOS

### LINKED ISSUE #353: DI framework for iOS targets
## Kind (Required)
- Proposal

 ## Overview (Required)
Why not use a DI framework for iOS targets?
Koin and Kodein are compatible with KMM.
As it is now, I think Hilt is fine for Android target. 
Because this app has to learn to experience Hilt as well.

 ## Links	
- https://insert-koin.io
- https://github.com/Kodein-Framework/Kodein-DI

watanavex:
## Issue
- close #353

## Overview (Required)
- This is a continuation of #396
- Firebase Auth abstracts and injects from Swift code
  - I was unable to link FirebaseAuth from the CocoaPods plugin in gradle.
  - GitLiveApp/firebase-kotlin-sdk uses Carthage
  - However, Firebase may discontinue support for Carthage.
  - So I thought of a way to abstract Firebase Auth and inject it from Swift code
- If you have another idea, it's welcome!

## Links
- https://github.com/GitLiveApp/firebase-kotlin-sdk/issues/133
- https://github.com/firebase/firebase-ios-sdk/blob/master/Carthage.md

## Screenshot
🙅‍♂️


takahirom:
👀

ios-framework/build.gradle.kts
```diff
@@ -0,0 +1,69 @@
...
+    val iosTarget: (String, KotlinNativeTarget.() -> Unit) -> KotlinNativeTarget =
+        if (System.getenv("SDK_NAME")?.startsWith("iphoneos") == true)
+            ::iosArm64
+        else
+            ::iosX64
+
+    iosTarget("ios") {
+        binaries {
+            framework {
+                baseName = "Shared"
```
takahirom, line 23: Thank you for creating iOS combine framework! 👍 
How about "DroidKaigiMPP" for framework name 👀

takahirom:
Will we implement Authenticator on the iOS side in the future? 👀

watanavex:
> Will we implement Authenticator on the iOS side in the future? 👀

Exactly.
Sorry for not explaining enough.
We will call this method from Swift.
As a parameter, we pass the Authenticator implemented in Swift.

https://github.com/DroidKaigi/conference-app-2021/blob/484e986ba39a23650e1a1e0729509a6af67a0117/ios-framework/src/iosMain/kotlin/io/github/droidkaigi/feeder/IosModule.kt#L10-L21

ios-framework/build.gradle.kts
watanavex: Sorry, I changed my name.
I'll change it back!

https://github.com/DroidKaigi/conference-app-2021/blob/484e986ba39a23650e1a1e0729509a6af67a0117/ios/project.yml#L52-L53

watanavex:
to-do
- Coroutines is locked to single-threaded version; change to multi-threaded version because Ktor doesn't work
- Implement Suspend Wrapper.
- Implement Authenticator in Swift

I'm expecting some problems with multithreading... :-x

ios-framework/build.gradle.kts
watanavex: Fixed 56e3f27 🙇‍♂️

takahirom (approved):
LGTM 👍 Thank you for advancing DroidKaigi!!

PR #416, opened 2021-04-18
remove ALL filter

### LINKED ISSUE #412: Remove `All` toggle
## Kind (Required)

- Improvement

 ## Overview (Required)

- Because it is different from the behavior of the check box.
All toggles are not needed. It exists in Figma, but the designer just can't erase it because she doesn't have time. 🙏 
image

 ## Links	

-

ked4ma:
## Issue
- close #412 

## Overview (Required)
- checkbox put on filter is not fit as checkbox behavior.
- so remove ALL filter from filter parts.

## Links
-

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/uicomponent-compose/feed/src/main/java/io/github/droidkaigi/feeder/feed/BackLayerContent.kt
+++ b/uicomponent-compose/feed/src/main/java/io/github/droidkaigi/feeder/feed/BackLayerContent.kt
@@ -4,17 +4,11 @@ import androidx.compose.foundation.background
 import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
-import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
-import androidx.compose.foundation.layout.width
-import androidx.compose.material.Icon
 import androidx.compose.material.MaterialTheme
 import androidx.compose.material.Surface
 import androidx.compose.material.Text
-import androidx.compose.material.icons.Icons
-import androidx.compose.material.icons.rounded.CheckBox
-import androidx.compose.material.icons.rounded.CheckBoxOutlineBlank
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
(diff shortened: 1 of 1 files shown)


uicomponent-compose/feed/src/main/java/io/github/droidkaigi/feeder/feed/BackLayerContent.kt
```diff
@@ -70,24 +64,14 @@ private fun Input(
                     Color.Transparent
                 }
             )
-            .padding(all = 12.dp)
+            .padding(horizontal = 20.dp, vertical = 12.dp)
```
ked4ma, line 67: when remove spacers (L83, L90), it seems to go left too much.
so i add additional padding here.

takahirom:
I'm sorry. 🙇 
Removing the "All" Filter means designing it as follows. 
image

ked4ma:
oh, i understood the spec of this issue.
i'll modify it.

ked4ma:
@takahirom 
i removed ONLY all filter, and restored about favorite filter!

uicomponent-compose/feed/src/main/java/io/github/droidkaigi/feeder/feed/BackLayerContent.kt
```diff
@@ -63,13 +55,6 @@ private fun Input(
         Modifier
             .clickable(onClick = onClick)
             .fillMaxWidth()
-            .background(
-                if (isChecked) {
-                    ConferenceAppFeederTheme.filterMuskColor
-                } else {
-                    Color.Transparent
-                }
-            )
```
takahirom, line 72: Thank you for applying the changes!
Is there any reason to remove the background? 👀

ked4ma: i thought this background musk is used to show which filter is applied.
and by this change, there are only ONE filter item and by checkbox, we can recognize whether favorite filter is selected or not.
so, i changed as above.
is it better to rollback about backgound ?

takahirom: I take a look at the material design guideline for a moment. :eyes:

takahirom: Certainly, it's not in the guidelines either. Let's erase it once. :+1:

takahirom (approved):
LGTM!

uicomponent-compose/feed/src/main/java/io/github/droidkaigi/feeder/feed/BackLayerContent.kt
ked4ma: thank you for your check!!!

Issue #344, opened 2021-03-11
The splash screen is not referencing the in-app theme


ocuto:
**Overview (Required)**
- The splash screen is not referencing the in-app theme. 
- Device settings will be referenced.

**Steps To Reproduce**
prerequire: System theme is light.

1. Go to 'Setting Screen'
2. Tap on 'Theme Column'
3. Select on 'Dark'
4.Go to 'Splash Screen'

**Expected behavior**
Become a 'Dark' theme

**Screenshots**
![before](https://user-images.githubusercontent.com/34264859/110724365-ff7c4800-8258-11eb-98dd-652741a97dce.gif)

**Env:**
 - Device: Pixel4
 - OS: 11.0

**Additional context**
ref: https://github.com/DroidKaigi/conference-app-2021/pull/339

PR #428, opened 2021-05-01
Fix #425: using setDefaultNightMode

### LINKED ISSUE #425: Drawable-night is not accessible
**Overview (Required)**
- In the dark mode, images are not coming from the `drawable-nigth`. 

**Steps To Reproduce**

1. Go to Settings and change theme to dark
2. Go to about and see the image logo

**Expected behavior**
In the dark theme image should come from `drawable-nigth`

**Screenshots**

**Env:**
 - Device: Pixel XL
 - OS: @* API level

**Additional context**
Approaches solve:
1. We can introduce `RealAboutThisAppViewModel` and `AboutThisAppViewModel` and bring the ThemeRepository data to the `AboutThisApp` screen and use the view model in the `fun WhatIsDroidKaigiComponent` and based on theme, get the image. But, in this approach, we cannot access an image from `drawable-night`, so we need to rename the image to `logo_night` and put it in the drawable folder only. 
2. We can use `ContextCompat` and get the drawable, but then we need to get the drawable id and pass it to the `painterResource` which is a bit complex.

anandwana001:
## Issue
- close #425

## Overview (Required)
- Using of setDefaultNightMode

## Links
-

## Screenshot
Before | After
:--: | :--:
 | 

This also solving #344

### DIFF
--- a/android/src/main/java/io/github/droidkaigi/feeder/MainActivity.kt
+++ b/android/src/main/java/io/github/droidkaigi/feeder/MainActivity.kt
@@ -2,17 +2,36 @@ package io.github.droidkaigi.feeder
 
 import android.os.Bundle
 import androidx.activity.compose.setContent
+import androidx.activity.viewModels
 import androidx.appcompat.app.AppCompatActivity
+import androidx.appcompat.app.AppCompatDelegate
 import androidx.core.view.WindowCompat
+import androidx.lifecycle.lifecycleScope
 import dagger.hilt.android.AndroidEntryPoint
 import io.github.droidkaigi.feeder.viewmodel.ProvideViewModels
+import io.github.droidkaigi.feeder.viewmodel.RealAppViewModel
+import kotlinx.coroutines.flow.launchIn
+import kotlinx.coroutines.flow.onEach
 
 @AndroidEntryPoint
 class MainActivity : AppCompatActivity() {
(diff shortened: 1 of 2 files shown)


anandwana001:
@takahirom Please let me know, if this is a correct approach to fix this issue.

takahirom:
I think there are good and bad things, but since the ViewModel will be in the form of directly operating the View, how about doing the following?

```kotlin
class MainActivity : AppCompatActivity() {
    private val appViewModel by viewModels<RealAppViewModel>()
    override fun onCreate(savedInstanceState: Bundle?) {
        super.onCreate(savedInstanceState)
        WindowCompat.setDecorFitsSystemWindows(window, false)
        setup()
        appViewModel.state
            .onEach {
// ...
            }
            .launchIn(lifecycleScope)
    }
```

anandwana001:
Using `setDefaultNightMode` will lead into activity recreation - https://medium.com/androiddevelopers/appcompat-v23-2-daynight-d10f90c83e94

takahirom:
I'll take a look tomorrow :eyes:

takahirom:
Now Activity is receated by config changes such as screen rotation. So that's fine.
Thank you for teaching me. 👍

takahirom (approved):
Thank you for raising the issue and solving it! 👍

PR #466, opened 2021-06-03
[iOS] Access color defined in Asset with Color directly

ostk0069:
## Issue
- close #461

## Overview (Required)
- able to use asset color with type Color which can use in SwiftUI directly.

## Links
-

## Screenshot
Before | After
:--: | :--:
 |

PR #467, opened 2021-06-04
[iOS] Support UIColor for ColorAsset

myihsan:
## Issue

#466 has accidentally broken the support of UIColor.

## Overview (Required)

- Support `UIColor` for `ColorAsset`
- Use `callAsFunction` to overload colors
- Update codes which using `AssetColor`

## Proposal

If this approach is fine, I would like to apply it to `ImageAsset`.

## Another Plan

#472

### DIFF
--- a/ios/Sources/AppFeature/AppScreen.swift
+++ b/ios/Sources/AppFeature/AppScreen.swift
@@ -64,9 +64,9 @@ public struct AppScreen: View {
 
     public init(store: Store<AppState, AppAction>) {
         self.store = store
-        UITabBar.appearance().barTintColor = AssetColor.Background.contents.color
-        UITabBar.appearance().unselectedItemTintColor = AssetColor.Base.disable.color
-        UINavigationBar.appearance().barTintColor = AssetColor.Background.primary.color
+        UITabBar.appearance().barTintColor = AssetColor.Background.contents()
+        UITabBar.appearance().unselectedItemTintColor = AssetColor.Base.disable()
+        UINavigationBar.appearance().barTintColor = AssetColor.Background.primary()
     }
 
     public var body: some View {
@@ -83,8 +83,8 @@ public struct AppScreen: View {
                 }
             }
         )
-        .accentColor(Color(AssetColor.primary.color.cgColor))
-        .background(Color(AssetColor.Background.primary.color))
+        .accentColor(AssetColor.primary())
+        .background(AssetColor.Background.primary())
     }
 }
 
--- a/ios/Sources/Component/Card/LargeCard.swift
+++ b/ios/Sources/Component/Card/LargeCard.swift
@@ -36,7 +36,7 @@ public struct LargeCard: View {
 
                 Text(title)
                     .font(.headline)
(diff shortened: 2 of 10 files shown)


ostk0069:
@myihsan 
Thanks for Contribution! & very sorry for having error with my PR... 😢  

I suggest to use Color Assets like this.
- `AssetColor.primaryDark.color` stands for `SwiftUI.Color`
- `AssetColor.primaryDark.uiColor` stands for `UIColor`

`CallAsFunction` is a great approach but in my opinion, this  looks better to read.
If you are okay with this, could you fix like this?

myihsan:
@ostk0069 

Thank you for your feedback!

> Asset**Color**.primaryDark.**color**

It's a little repetitive to me.
But I am totally okay without using `callAsFunction`.

> AssetColor.primaryDark.uiColor stands for UIColor

Did you mean we should drop the support for `NSColor` or switch to different approach?
With current approach, `colorObject` may be just fine?

ostk0069:
@myihsan

> It's a little repetitive to me.

I understand your opinion. Therefore, this `AssetColor` stands for the file name `color.xcassets`.
I think there is also benefit to divide the file of images and colors.

> Did you mean we should drop the support for NSColor or switch to different approach?
With current approach, colorObject may be just fine?

I don't think my idea mean it does not support NSColor. 
I meant like this. 
- if `macOS` -> `color` stands for `NSColor` 
- if `iOS(equal or higher than 13.0)` -> `color` stands for `SwiftUI.Color` .

In addition, add `uiColor` extension when platform is `iOS(equal or higher than 11.0)`

myihsan:
@ostk0069 

> - if macOS -> color stands for NSColor
> - if iOS(equal or higher than 13.0) -> color stands for SwiftUI.Color
>
> In addition, add uiColor extension when platform is iOS(equal or higher than 11.0) 

`SwiftUI.Color` is also available in macOS 10.15+, so I can't tell why to provide it only in iOS.
Instead, I think it would be better to combine `UIColor` and `NSColor` as `ColorClass`, and provide `SwiftUI.Color` additionally, which is the same as the current implementation of this PR.

If there is any reason better to combine `NSColor` and `SwiftUI.Color`, please tell me.

ostk0069:
@myihsan 

> SwiftUI.Color is also available in macOS 10.15+, so I can't tell why to provide it only in iOS.
Instead, I think it would be better to combine UIColor and NSColor as ColorClass, and provide SwiftUI.Color additionally, which is the same as the current implementation of this PR.

Oh, you're right. Let's not combine NSColor and SwiftUI.Color 🙇 
In this project most of the time we are using is SwiftUI.Color. 
So, I suggest to name `SwiftUI.Color` as `color`. why don't name `UIColor` as `uiColor` and `NSColor` as `nsColor`

myihsan:
@ostk0069 
> So, I suggest to name SwiftUI.Color as color. why don't name UIColor as uiColor and NSColor as nsColor

It's fine to me.
But it needs a lot of changes from this PR, so I made another one #472.

Issue #445, opened 2021-05-31
What we need with Kotlin Multiplatform for iOS


takahirom:
## Kind (Required)

- Proposal / Discussion

 ## Overview (Required)

 - I want to sort out what we need a little.

according to https://github.com/DroidKaigi/conference-app-2021/pull/415#issuecomment-824710629

- Implement Suspend Wrapper.
https://touchlab.co/kotlin-coroutines-rxswift/

- Implement Authenticator in Swift

Authenticator

```kotlin
package io.github.droidkaigi.feeder

data class User(
    val idToken: String?
)

interface Authenticator {
    suspend fun currentUser(): User?
    suspend fun signInAnonymously(): User?
}
```

 ## Links	

-

takahirom:
@watanavex 
Is this the same recognition? 🙏

watanavex:
@takahirom  Thank you!
- Implement Suspend Wrapper
[Suspend wrapper has already been implemented.](https://github.com/DroidKaigi/conference-app-2021/tree/main/ios-framework/src/iosMain/kotlin/io/github/droidkaigi/feeder/repository)
This is a good match with the README described by @ry-itto, and I am aware of no problems.

- Implement Authenticator in Swift
We need this. I want to do it!!

watanavex:
@takahirom 
How do I get the GoogleService-Info.plist?
(In the context of Android, it is google-services.json)

takahirom:
👀

takahirom:
I'm checking it, so please wait for a while.

takahirom:
@watanavex 
Please use this 🙏 
[GoogleService-Info-Stating-Debug.plist.zip](https://github.com/DroidKaigi/conference-app-2021/files/6602154/GoogleService-Info-Stating-Debug.plist.zip)

PR #469, opened 2021-06-04
[iOS] Create xcframework with Gradle

### LINKED ISSUE #459: [iOS & KMM] Generate KMM module as xcframework
## Kind (Required)

- Proposal / Discussion
- Improvement

 ## Overview (Required)

- Currently, the KMM module is generated as `.framework`. In iOS App, use `xcframework` is better way. 
- In this iOS project management way, it can use `xcframework` only.
- According to above, it is necessary to generate KMM module as `xcframework`. 

There are two solutions.
1. Generate `xcframework` with gradle step.
  - It's not clear, but I think this plugin may be useful.
    - https://github.com/ge-org/multiplatform-swiftpackage
2. Generate `xcframework` with shell script.
  - Using generated `.framework` and combine them.

I want to decide the generating way. 🙏🏼 

 ## Links	

This discussion may be related. (Cocoapods part is not needed.)
- https://discuss.kotlinlang.org/t/distribute-kmm-shared-as-xcframework/20163

ry-itto:
## Issue
- close #459 

## Overview (Required)
- Create XCFrameworks of KMM module with gradle.
  - @takahirom told me about this  https://github.com/DroidKaigi/conference-app-2021/issues/459#issuecomment-853481869, but I couldn't create with that. 😢 
  - So I used plugin.
- Add target in Package.swift about KMM module

## Links
-

## Screenshot
Before | After
:--: | :--:
 |


ios-framework/build.gradle.kts
```diff
@@ -75,3 +76,11 @@ tasks.getByName("build").dependsOn(packForXcode)
 
 // Workaround for issues where types defined in iOS native code cannot be referenced in Android Studio
 tasks.getByName("preBuild").dependsOn(tasks.getByName("compileKotlinIos"))
+
+multiplatformSwiftPackage {
+    swiftToolsVersion("5.3")
+    targetPlatforms {
+        iOS { v("14") }
+    }
+    outputDirectory(File(projectDir, "../ios/build/xcframeworks"))
```
ry-itto, line 85: SPM can't refer outside of Swift Package binary, so place built `XCFramework` to `ios/build/xcframeworks`

ry-itto:
### Question
- Build KMM module with `iosArm64` architecture doesn't work. Is it an expected behavior?
  - I tried to build `iosArm64` architecture framework with following.
  Change this

  ```
    val iosTarget: (String, KotlinNativeTarget.() -> Unit) -> KotlinNativeTarget =
        if (System.getenv("SDK_NAME")?.startsWith("iphoneos") == true)
            ::iosArm64
        else
            ::iosX64
  ```
  to
  ```
  val iosTarget: (String, KotlinNativeTarget.() -> Unit) -> KotlinNativeTarget = ::iosArm64
  ```

- I think the KMM module isn't currently available from iOS. Is it true?
  - When I tried to use `**Repository` in KMM module, it's concrete class can't instantiate. The class constructor's argument requires `Api` interface, but no implemented classes are available.
  - Will it be available after https://github.com/DroidKaigi/conference-app-2021/issues/445 implemented?

@takahirom @watanavex 
I would appreciate it if you could answer 🙏🏼

takahirom:
I will check it🙏

takahirom:
I think this SDK_NAME is an environment variable that is added when it is executed in the build phase of XCode. Could you check it? 👀 
https://matsukaz.hatenablog.com/entry/2015/07/06/090933

takahirom:
`println("SDK_NAME:"+System.getenv("SDK_NAME"))`

I think you can check the environment variable in build.gradle file.

ry-itto:
> I think this SDK_NAME is an environment variable that is added when it is executed in the build phase of XCode. Could you check it? 👀
> https://matsukaz.hatenablog.com/entry/2015/07/06/090933

Yes! It's true.
I tried to call `./gradlew ios-framework:createXCFramework` with terminal so `SDK_NAME` is none.

What I wanted to do here

> Build KMM module with iosArm64 architecture doesn't work. Is it an expected behavior?

is creating framework for `iosArm64` (for real device) is failed.

This means currently I think that it may not be possible to call KMM on the real device.

The result of executing `./gradlew ios-framework:createXCFramework` is following 🙏🏼 
[gradle-createXCFramework.log](https://github.com/DroidKaigi/conference-app-2021/files/6602187/gradle-createXCFramework.log)

takahirom:
Thank you! I am looking 👀 

```
         project :ios-framework
      > No matching variant of project :data:repository was found. The consumer was configured to find a usage of 'kotlin-api' of a component, as well as attribute 'org.jetbrains.kotlin.platform.type' with value 'native', attribute 'org.jetbrains.kotlin.native.target' with value 'ios_arm64' but:
          - Variant 'androidDebugAndroidTestCompile' capability conference-app-2021.data:repository:unspecified:
              - Incompatible because this component declares a component, as well as attribute 
```

takahirom:
Could you try this?

```
export SDK_NAME=iphoneos
./gradlew ios-framework:createXCFramework
```

If you are doing the following, it will fail because there is a similar if statement in repository etc.

`val iosTarget: (String, KotlinNativeTarget.() -> Unit) -> KotlinNativeTarget = ::iosArm64`

https://github.com/DroidKaigi/conference-app-2021/pull/469#issuecomment-854914329


<details>

```
The Kotlin source set androidAndroidTestRelease was configured but not added to any Kotlin compilation. You can add a source set to a target's compilation by connecting it with the compilation's default source set using 'dependsOn'.
See https://kotlinlang.org/docs/reference/building-mpp-with-gradle.html#connecting-source-sets

> Task :data:db:generateIosMainDatabaseInterface

> Task :model:compileKotlinIosArm64

> Task :data:db:compileKotlinIos

> Task :data:api:compileKotlinIos

> Task :data:repository:compileKotlinIos
> Task :ios-framework:compileKotlinIos
> Task :ios-framework:linkReleaseFrameworkIos

> Task :ios-framework:createXCFramework
xcframework successfully written out to: /Users/takahirom/git/conference-app-2021/ios/build/xcframeworks/DroidKaigiMPP.xcframework

Deprecated Gradle features were used in this build, making it incompatible with Gradle 8.0.
Use '--warning-mode all' to show the individual deprecation warnings.
See https://docs.gradle.org/7.0/userguide/command_line_interface.html#sec:command_line_warnings

BUILD SUCCESSFUL in 2m 20s
```
</details>

ry-itto:
@takahirom 
> If you are doing the following, it will fail because there is a similar if statement in repository etc.

I understand!
I missed 🙏🏼 

Thank you! The task succeeded!

myihsan:
@ry-itto 
Did you able to build the XCFramework, which contains frameworks for ARM64 and X64?
If not, I would like to submit a PR to your branch to show my solution.
It seems like we need to update the workaround related to the hierarchical structure.

> * I think the KMM module isn't currently available from iOS. Is it true?
>   
>   * When I tried to use `**Repository` in KMM module, it's concrete class can't instantiate. The class constructor's argument requires `Api` interface, but no implemented classes are available.
>   * Will it be available after #445 implemented?

https://github.com/DroidKaigi/conference-app-2021/blob/main/ios-framework/src/iosMain/kotlin/io/github/droidkaigi/feeder/IosModule.kt

You should be able to use the object resolved by Koin.

ry-itto:
> Did you able to build the XCFramework, which contains frameworks for ARM64 and X64?

No, I can't.
Currently, build target architecture only 😢 

> If not, I would like to submit a PR to your branch to show my solution.

Yes, please! 🙏🏼

myihsan:
@ry-itto 
Please refer to https://github.com/ry-itto/conference-app-2021/pull/1

@takahirom @watanavex 
If there is any better solution, please tell me.

takahirom:
Does this approach result in the iosMain directory being an arm64 directory instead of being shared? 👀 

Some of the approaches I know are:

```groovy
    if (System.getProperty("idea.active") == "true") {
        // Fix IntelliJ Unresolved reference
        def onPhone = System.getenv("SDK_NAME")?.startsWith("iphoneos") ?: false
        if (onPhone) {
            iosArm64("ios")
        } else {
            iosX64("ios")
        }
    } else {
        ios()
    }
```

This is not so good because the environment changes depending on whether it is executed from the terminal or from Idea.

takahirom:
@myihsan

Does this approach result in the iosMain directory being an arm64 directory instead of being shared? 👀 

Some of the approaches I know are:

```groovy
    if (System.getProperty("idea.active") == "true") {
        // Fix IntelliJ Unresolved reference
        def onPhone = System.getenv("SDK_NAME")?.startsWith("iphoneos") ?: false
        if (onPhone) {
            iosArm64("ios")
        } else {
            iosX64("ios")
        }
    } else {
        ios()
    }
```

This is not so good because the environment changes depending on whether it is executed from the terminal or from Idea.

takahirom:
Maybe if you use this, you can go only with `ios()`. .. It may not be realistic as it takes time to investigate.
https://github.com/LouisCAD/CompleteKotlin

myihsan:
> Does this approach result in the iosMain directory being an arm64 directory instead of being shared? 👀

I am not sure what you mean, but it's being shared by `iosX64Main.dependsOn(it)`.
And the product (XCFramework) can be run both on a simulator and a real device, so it seems no problem.

The solution you provided can cover most cases, but how about run the iOS app from Android Studio using KMM plugin?

watanavex:
Sorry, I'm late checking this.
I'll catch up soon. 👀

takahirom:
@myihsan 
Thank you! 
I missed `iosX64Main.dependsOn(it)` 👀 
I didn't know it would work this way.

It's important to be able to do it for the time being, so I think it's good! 👍 
I think it's better to do just one, how about fixing it as follows?
I'm afraid that if one source set makes changes to another source set, it will be difficult to find the correct information.

```kotlin
        val iosX64Main by getting
        val iosMain by getting {
            dependencies {
                implementation(Dep.Koin.core)
            }
            iosX64Main.dependsOn(this)
        }
```

↓

```kotlin
        val iosMain by getting {
            dependencies {
                implementation(Dep.Koin.core)
            }
        }
        val iosX64Main by getting {
           dependsOn(iosMain)
        }
```

myihsan:
@takahirom 
You are right! I have followed the document of hierarchical structure without enough thought.
I will fix it.

watanavex:
@myihsan 
I have no better ideas.
and I've learned a lot from your idea 👏

ry-itto:
@myihsan 
Please tell me if the PR is ready to review 🙏🏼 

Should @takahirom review for this PR (https://github.com/ry-itto/conference-app-2021/pull/1) and approve this as well?

myihsan:
@ry-itto 
It's ready! I fixed `build.gradle` as takahirom has told me.

takahirom (approved):
LGTM 👍 Thank you for your challenge to Kotlin Multiplatform!

ry-itto:
@takahirom 
Thank you for your review!
But the CI task ( `test-ios` ) failed still remains 😢 

Please help about that?

It seems like a gradle task `iosTest` is ambiguous.

https://github.com/DroidKaigi/conference-app-2021/runs/2753974134?check_suite_focus=true

takahirom:
It seems that GitHub Actons does not have iosTest settings anymore, so I thought that CI settings might not be reflected unless merged.

ry-itto:
It occurs my local environment too...

```
FAILURE: Build failed with an exception.

* What went wrong:
Task 'iosTest' is ambiguous in root project 'conference-app-2021'. Candidates are: 'iosArm64TestBinaries', 'iosArm64TestKlibrary', 'iosArm64TestProcessResources', 'iosTestBinaries', 'iosTestKlibrary', 'iosTestProcessResources', 'iosX64Test', 'iosX64TestBinaries', 'iosX64TestKlibrary', 'iosX64TestProcessResources'.
```

takahirom:
I will check it 👀

takahirom:
The task called iosTest is gone, and this PR removes it from the CI settings.
What task did this error run?
https://github.com/DroidKaigi/conference-app-2021/pull/469#issuecomment-855348465

ry-itto:
The CI setting of ios-test originally renamed with test-ios in https://github.com/DroidKaigi/conference-app-2021/pull/471 .

Currently, renamed originally CI setting is failing.
The task is still iosTest.

This PR's change can be ignored.

takahirom:
I think that the CI settings will be reflected if they are merged.
How about reverting if it doesn't work? 👀

ry-itto:
ok, I will merge this 🙏🏼

PR #473, opened 2021-06-05
[iOS] Fix previous color properties

woxtu:
## Issue
- close N/A

## Overview (Required)
- It seems like the latest commit cannot build, so this PR contains the following changes to fix that:
  - Initialize all properties in initializer to fix a compile error
  - Change the type of previous color properties to avoid redundant conversions

## Links
-

## Screenshot
Before | After
:--: | :--:
 |


ry-itto (approved):
lgtm 💯 
Thanks!

PR #484, opened 2021-06-06
[iOS] Change default CPU architecture

myihsan:
## Issue

Fix #469 

## Overview (Required)

It seems like `iosArm64` doesn't have a test task and `iosX64Test` can't depend on `iosTest`, which builds `iosArm64`.
So change `ios` to `iosX64` should work.

I have chosen`iosArm64` as the default CPU architecture since it's more suitable to call `ios`, but it looks like a bad idea.


takahirom (approved):
LGTM 👍 Thank you for your support!

takahirom:
@myihsan The PR was drafted, can I merge it?

myihsan:
@takahirom 
Yes, please!
I left it as a draft to wait for the test result but forgot to make it ready.

PR #489, opened 2021-06-09
Add KMM Swift Models

ry-itto:
## Issue
- close none

## Overview (Required)
- Create KMM Swift Models
- [x] Add `Model` module build in CI

## Links
-

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/ios/Sources/Component/LoadingStatus.swift
+++ b/ios/Sources/Component/LoadingStatus.swift
@@ -1,0 +1,6 @@
+public enum LoadingStatus<T> {
+    case initial
+    case loading
+    case loaded(T)
+    case error(Error)
+}
--- a/ios/Sources/Model/Blog.swift
+++ b/ios/Sources/Model/Blog.swift
@@ -1,0 +1,56 @@
+import DroidKaigiMPP
+
+@dynamicMemberLookup
+public struct Blog: Equatable, Identifiable {
+    public var feedItem: FeedItem
+    public var author: Author
+    public var language: String
+
+    public var id: String {
+        feedItem.id
+    }
+
+    public init(
+        id: String,
+        image: Image,
+        link: String,
+        media: Media,
+        publishedAt: Date,
+        summary: MultiLangText,
+        title: MultiLangText,
+        author: Author,
+        language: String
+    ) {
+        self.feedItem = .init(
+            id: id,
+            image: image,
+            link: link,
+            media: media,
+            publishedAt: publishedAt,
+            summary: summary,
+            title: title
+        )
+        self.author = author
+        self.language = language
+    }
+
+    public init(from model: DroidKaigiMPP.FeedItem.Blog) {
+        self.feedItem = .init(
+            id: model.id,
+            image: Image(from: model.image),
+            link: model.link,
+            media: Media.from(model.media),
+            publishedAt: Date(
+                timeIntervalSince1970: Double(model.publishedAt.toEpochMilliseconds())
+            ),
+            summary: MultiLangText(from: model.summary),
+            title: MultiLangText(from: model.title)
+        )
+        self.author = Author(from: model.author)
+        self.language = model.language
+    }
+
+    public subscript<T>(dynamicMember keyPath: KeyPath<FeedItem, T>) -> T {
+        self.feedItem[keyPath: keyPath]
+    }
+}
--- a/ios/Sources/Model/FeedContents.swift
+++ b/ios/Sources/Model/FeedContents.swift
@@ -1,0 +1,17 @@
+import DroidKaigiMPP
+
+public struct FeedContents: Equatable {
+    public struct Content: Equatable {
+        public var item: FeedItemType
+        public var isFavorited: Bool
+
+        public init(item: FeedItemType, isFavorited: Bool) {
+            self.item = item
+            self.isFavorited = isFavorited
+        }
+    }
+
+    public var contents: [Content]
+    public var favorites: Set<String>
+    public var feedItemContents: [FeedItemType]
+}
--- a/ios/Sources/Model/FeedContents.swift
+++ b/ios/Sources/Model/FeedContents.swift
@@ -2,16 +2,37 @@ import DroidKaigiMPP
 
 public struct FeedContents: Equatable {
     public struct Content: Equatable {
-        public var item: FeedItemType
+        public var item: FeedItemType?
         public var isFavorited: Bool
 
         public init(item: FeedItemType, isFavorited: Bool) {
             self.item = item
             self.isFavorited = isFavorited
         }
+
+        public init(from model: DroidKaigiMPP.KotlinPair<DroidKaigiMPP.FeedItem, KotlinBoolean>) {
+            self.item = model.first.flatMap(FeedItemType.from(_:))
+            self.isFavorited = model.second?.boolValue ?? false
+        }
     }
 
     public var contents: [Content]
     public var favorites: Set<String>
     public var feedItemContents: [FeedItemType]
+
+    public init(
+        contents: [Content],
+        favorites: Set<String>,
+        feedItemContents: [FeedItemType]
+    ) {
+        self.contents = contents
+        self.favorites = favorites
+        self.feedItemContents = feedItemContents
+    }
+
+    public init(from model: DroidKaigiMPP.FeedContents) {
+        self.contents = model.contents.map(Content.init(from:))
+        self.favorites = model.favorites
+        self.feedItemContents = model.feedItemContents.compactMap(FeedItemType.from(_:))
+    }
 }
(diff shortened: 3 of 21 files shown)


ios/Sources/Model/Blog.swift
```diff
@@ -0,0 +1,56 @@
+import DroidKaigiMPP
+
+@dynamicMemberLookup
```
ostk0069, line 3: Just in case, this is fine with this, right?

ios/Sources/Model/FeedContents.swift
```diff
@@ -0,0 +1,17 @@
...
+        public var isFavorited: Bool
+
+        public init(item: FeedItemType, isFavorited: Bool) {
+            self.item = item
+            self.isFavorited = isFavorited
+        }
+    }
+
+    public var contents: [Content]
+    public var favorites: Set<String>
```
ostk0069, line 15: Just in case, this is fine with this, right?

ios/Sources/Model/MultiLangText.swift
```diff
@@ -0,0 +1,32 @@
...
+        self.currentLangTitle = currentLangTitle
+    }
+
+    public init(from model: DroidKaigiMPP.MultiLangText) {
+        self.enTitle = model.enTitle
+        self.jaTitle = model.jaTitle
+        self.currentLangTitle = model.currentLangTitle
+    }
+
+    public func getByLang(lang: Lang) -> String {
```
ostk0069, line 24: 👍

ios/Sources/Model/Image.swift
```diff
@@ -0,0 +1,23 @@
...
+        self.largeURLString = largeURLString
+        self.smallURLString = smallURLString
+        self.standardURLString = standardURLString
+    }
+
+    public init(from model: DroidKaigiMPP.Image) {
+        self.largeURLString = model.largeUrl
+        self.smallURLString = model.smallUrl
+        self.standardURLString = model.standardUrl
+    }
```
ostk0069, line 22: Better creating enum and function to take out? which is like
```
public func getByLang(lang: Lang) -> String
```

ios/Sources/Component/LoadingStatus.swift
```diff
@@ -0,0 +1,6 @@
+public enum LoadingStatus<T> {
```
ostk0069, line 1: Looks like `loaded` could remind me of meaning of error as well.
In my opinion, `pending`, `loading(or running)`, `success`, `failure` might be better for names. 
ref: https://github.com/ra1028/SwiftUI-Hooks/blob/main/Sources/Hooks/AsyncStatus.swift

ostk0069 (commented):
Just let me know that replacing current dummy model will be done in this PR or not?

ry-itto:
> Just let me know that replacing current dummy model will be done in this PR or not?
No, that will be fixed in feature implementation tasks

ios/Sources/Model/FeedContents.swift
ry-itto: Basically, this PR's definitions are based on KMM module.
So I think this is fine.

ios/Sources/Model/Image.swift
ry-itto: I think this is not needed. If I do it, I will have to define `ImageSize` enum and it's redundancy.

takahirom:
I'm not sure, but what maps to the Swift model is that it's difficult to use the KMM model in an iOS app? 👀

ios/Sources/Model/FeedItem.swift
```diff
@@ -0,0 +1,51 @@
...
+        self.publishedDateString = nil
+    }
+
+    public init(from model: DroidKaigiMPP.FeedItem) {
+        self.id = model.id
+        self.image = Image(from: model.image)
+        self.link = model.link
+        self.media = Media.from(model.media)
+        self.publishedAt = Date(
+            timeIntervalSince1970: Double(model.publishedAt.toEpochMilliseconds())
```
myihsan, line 44: badge
May be a better choice to [export](https://kotlinlang.org/docs/mpp-build-native-binaries.html#export-dependencies-to-binaries) kotlinx-datetime since it has [`Conversations`](https://github.com/Kotlin/kotlinx-datetime/blob/master/core/darwin/src/Converters.kt) which can help us to keep the consistency when we do something similar in `iosMain`?

myihsan:
> I'm not sure, but what maps to the Swift model is that it's difficult to use the KMM model in an iOS app? 👀

Not sure if it is difficult to use in any case, but it's difficult when using the Composable Architecture, which deeply depends on `Equatable`.

ry-itto:
> I'm not sure, but what maps to the Swift model is that it's difficult to use the KMM model in an iOS app? 👀

- Dependency perspective
In this iOS app, because of multi module targets, `**Feature` module shouldn't see `**Repository`, but by using Model that generated by KMM, `**Feature` module can see `**Repository` class (KMM Repository class) 😢 


> > I'm not sure, but what maps to the Swift model is that it's difficult to use the KMM model in an iOS app? 👀
> 
> Not sure if it is difficult to use in any case, but it's difficult when using the Composable Architecture, which deeply depends on `Equatable`.

@myihsan says reason too 🙏🏼 

So in this iOS app, converting Model from KMM Model .

ios/Sources/Model/FeedItem.swift
ry-itto: I'm not sure,

Does `export kotlinx-datetime` mean creating a converting function in KMM module that can be used in Swift? For example, `Instant.toNSDate()` can be used in Swift.

myihsan: > Does export kotlinx-datetime mean creating a converting function in KMM module that can be used in Swift? For example, Instant.toNSDate() can be used in Swift.

Yes! But strictly speaking, it's not creating but export those functions.

takahirom:
> Not sure if it is difficult to use in any case, but it's difficult when using the Composable Architecture, which deeply depends on Equatable.

It may be unavoidable in the current architecture that it is necessary to convert to a form suitable for the OS in places close to the UI such as ViewModel.

> In this iOS app, because of multi module targets, **Feature module shouldn't see **Repository, but by using Model that generated by KMM, **Feature module can see **Repository class (KMM Repository class) 😢

There is certainly a problem that the visibility of modules cannot be divided on iOS.
I think that it is possible to hide Reposotry if it is implemented in KMM to a place close to the UI such as ViewModel, but this may also be difficult with the current architecture. 👀 
I think you can do what you want. 👍

ios/Sources/Model/Media.swift
```diff
@@ -0,0 +1,24 @@
...
+    static func from(_ model: DroidKaigiMPP.Media) -> Media {
+        switch model {
+        case let fm as DroidKaigiMPP.Media.DroidKaigiFM:
+            return .droidkaigifm(fm.text)
+        case let medium as DroidKaigiMPP.Media.Medium:
+            return .medium(medium.text)
+        case let youtube as DroidKaigiMPP.Media.YouTube:
+            return .youtube(youtube.text)
+        case let other as DroidKaigiMPP.Media.Other:
+            return .other(other.text)
```
myihsan, line 18: badge
Is the `text` necessary for us?
It seems like the raw value for saving to the database.

ios/Sources/Model/MultiLangText.swift
```diff
@@ -0,0 +1,32 @@
+import DroidKaigiMPP
+
+public struct MultiLangText: Equatable {
+    public var enTitle: String
+    public var jaTitle: String
+    public var currentLangTitle: String
```
myihsan, line 6: badge
Why we need `currentLangTitle` here as a stored property?
In my opinion, to reflect the change of `Lang`, a `View` using `MultiLangText` can observe the change of `Lang` directly without `currentLangTitle` and get the proper value from `getByLang(lang:)`.

ios/Sources/Model/Media.swift
ry-itto: > It seems like the raw value for saving to the database.

I see. I didn't understand about it 😢 
I will remove it 🙏🏼 
If there is any inconvenience later, I will revert it.

ios/Sources/Model/MultiLangText.swift
ry-itto: Basically, the model converted from KMM model has equal properties.
But I agree your opinion, so I will remove this property.
If there is any inconvenience later, I will revert it too. 🙏🏼

ios/Sources/Model/FeedItem.swift
ry-itto: Sorry for late replying 🙇🏼 

Can you do it another PR after this PR merged?
I don't know how to do it 😭

ios/Sources/Component/LoadingStatus.swift
ostk0069: @ry-itto 
How about this one?

ry-itto: This `LoadingStatus` is equal to KMM module's `LoadingState`. So I think it shouldn't change to other name cases.

If we don't need to use KMM module's `LoadingState` in Model module scope, I think your opinion will be better.

ostk0069 (approved):

PR #497, opened 2021-06-13
[iOS] Export kotlinx-datetime to the iOS Framework

myihsan:
## Overview (Required)

To keep all conversion between `kotlinx-datetime` and Swift types consistently, export `kotlinx-datetime`'s  [`Converters.kt`](https://github.com/Kotlin/kotlinx-datetime/blob/master/core/darwin/src/Converters.kt) to iOS.

## Links

- https://github.com/Kotlin/kotlinx-datetime

### DIFF
--- a/ios-framework/build.gradle.kts
+++ b/ios-framework/build.gradle.kts
@@ -23,6 +23,7 @@ kotlin {
                 baseName = "DroidKaigiMPP"
                 export(project(":model"))
                 export(project(":data:repository"))
+                export(Dep.datetime)
                 linkerOpts.add("-lsqlite3")
             }
         }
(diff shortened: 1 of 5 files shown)


ry-itto:
Can't `toKotlinInstant` function be used ?
We need that function too 😭 

スクリーンショット 2021-06-13 11 49 02

https://github.com/Kotlin/kotlinx-datetime/blob/master/core/darwin/src/Converters.kt#L31

myihsan:
Yes, we can, but have to use it this way.
```swift
ConvertersKt.toKotlinInstant(Date())
```

ry-itto (approved):
LTGM 🙏🏼

PR #515, opened 2021-06-21
[iOS][Proposal] Refactor FeedItem to a protocol

myihsan:
## Issue

In the major cases, the sealed class in Kotlin should be treated as an enumeration in Swift to be exhaustive in `switch`.
However, for `FeedItem`, we need the common properties to display them in sequence, so we have to extract the associated value by `switch` every time we access a property if we treat it as an enumeration.

Therefore, it may be better to treat it as a protocol.

## Overview (Required)

- Refactor `FeedItem` to a protocol
- Use `AnyFeedItem` to wrap values of types implemented `FeedItem` in order to implement `Equatable` without losing the capability to use without generic (for example when `FeedItem` inherited `Equatable` ).

### DIFF
--- a/ios/Sources/Model/Blog.swift
+++ b/ios/Sources/Model/Blog.swift
@@ -47,8 +47,8 @@ public struct Blog: FeedItem, Equatable {
 }
 
 public extension Blog {
-    var kmmModel: DroidKaigiMPP.FeedItem {
-        DroidKaigiMPP.FeedItem.Blog(
+    var kmmModel: DroidKaigiMPP.FeedItem.Blog {
+        .init(
             id: id,
             publishedAt: ConvertersKt.toKotlinInstant(publishedAt),
             image: image.kmmModel,
@@ -60,4 +60,6 @@ public extension Blog {
             author: author.kmmModel
         )
     }
+
+    var _kmmModel: DroidKaigiMPP.FeedItem { kmmModel }  // swiftlint:disable:this identifier_name
 }
--- a/ios/Sources/Model/Blog.swift
+++ b/ios/Sources/Model/Blog.swift
@@ -1,14 +1,15 @@
 import DroidKaigiMPP
 
-@dynamicMemberLookup
-public struct Blog: Equatable, Identifiable {
-    public var feedItem: FeedItem
+public struct Blog: FeedItem, Equatable {
+    public var id: String
+    public var image: Image
+    public var link: String
+    public var media: Media
+    public var publishedAt: Date
+    public var summary: MultiLangText
+    public var title: MultiLangText
     public var author: Author
     public var language: String
-
-    public var id: String {
-        feedItem.id
-    }
 
     public init(
         id: String,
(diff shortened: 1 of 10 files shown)


ios/Sources/Model/FeedItem.swift
```diff
@@ -1,72 +1,52 @@
...
-        switch self {
-        case let .blog(blog):
-            return blog.kmmModel
-        case let .podcast(podcast):
-            return podcast.kmmModel
-        case let .video(video):
-            return video.kmmModel
-        }
+@dynamicMemberLookup
+public struct AnyFeedItem: Equatable {
```
myihsan, line 15: Better to name it `SomeFeedItem`?

ry-itto: `AnyFeedItem` is okey 👍🏼

ios/Sources/Model/Blog.swift
```diff
@@ -21,48 +22,40 @@ public struct Blog: Equatable, Identifiable {
...
-            summary: feedItem.summary.kmmModel,
-            link: feedItem.link,
+            publishedAt: ConvertersKt.toKotlinInstant(publishedAt),
+            image: image.kmmModel,
+            media: media.kmmModel,
+            title: title.kmmModel,
+            summary: summary.kmmModel,
+            link: link,
             language: language,
             author: author.kmmModel
```
ry-itto, line 60: Why this type changed to `DroidKaigiMPP.FeedItem` ?
I think this `kmmModel` should convert Swift model to appropriate KMM model.
So I think this type should be `DroidKaigiMPP.FeedItem.Blog`

ios/Sources/Model/Podcast.swift
```diff
@@ -21,48 +22,40 @@ public struct Podcast: Equatable, Identifiable {
...
-
-    public subscript<T>(dynamicMember keyPath: KeyPath<FeedItem, T>) -> T {
-        self.feedItem[keyPath: keyPath]
-    }
 }
 
 public extension Podcast {
-    var kmmModel: DroidKaigiMPP.FeedItem.Podcast {
-        .init(
+    var kmmModel: DroidKaigiMPP.FeedItem {
```
ry-itto, line 50: ditto

ios/Sources/Model/Video.swift
```diff
@@ -17,44 +18,36 @@ public struct Video: Equatable, Identifiable {
...
+        self.publishedAt = model.publishedAt.toNSDate()
+        self.summary = MultiLangText(from: model.summary)
+        self.title = MultiLangText(from: model.title)
     }
 }
 
 public extension Video {
-    var kmmModel: DroidKaigiMPP.FeedItem.Video {
-        .init(
+    var kmmModel: DroidKaigiMPP.FeedItem {
```
ry-itto, line 42: ditto

ios/Sources/Repository/FeedRepository.swift
```diff
@@ -33,9 +33,9 @@ public struct FeedRepository: FeedRepositoryProtocol, KMMRepositoryProtocol {
         .eraseToAnyPublisher()
     }
 
-    public func addFavorite(feedItem: Model.FeedItemType) -> AnyPublisher<Void, KotlinError> {
+    public func addFavorite(feedItem: AnyFeedItem) -> AnyPublisher<Void, KotlinError> {
         Future<Void, KotlinError> { promise in
-            repository.addFavorite(feedItem: feedItem.item)
+            repository.addFavorite(feedItem: feedItem.wrappedValue.kmmModel)
```
ry-itto, line 38: How about this?
```suggestion
            repository.addFavorite(feedItem: feedItem.kmmModel)
```

ios/Sources/Repository/FeedRepository.swift
```diff
@@ -45,9 +45,9 @@ public struct FeedRepository: FeedRepositoryProtocol, KMMRepositoryProtocol {
         }.eraseToAnyPublisher()
     }
 
-    public func removeFavorite(feedItem: Model.FeedItemType) -> AnyPublisher<Void, KotlinError> {
+    public func removeFavorite(feedItem: AnyFeedItem) -> AnyPublisher<Void, KotlinError> {
         Future<Void, KotlinError> { promise in
-            repository.removeFavorite(feedItem: feedItem.item)
+            repository.removeFavorite(feedItem: feedItem.wrappedValue.kmmModel)
```
ry-itto, line 50: ditto
```suggestion
            repository.removeFavorite(feedItem: feedItem.kmmModel)
```

ry-itto (commented):
LGTM 💯 
Thank you for cleaning code 🙏🏼

ios/Sources/Model/Blog.swift
myihsan: I adjusted the interface to achieve your request. ab019c8

And I would also like to ask if we actually need `kmmModel`.
It's required by `FeedRepository`, but the only information it needs is `id` according to the implementation.
https://github.com/DroidKaigi/conference-app-2021/blob/be8c9452d13ef5b6e9121553f2c3027ce53f647a/data/repository/src/commonMain/kotlin/io/github/droidkaigi/feeder/data/FeedRepositoryImpl.kt#L27-L33
So we may adjust the API of `FeedRepository` only to require an `id`, instead of introducing `kmmModel`.
Or if there any other reason that requires `kmmModel`?

ry-itto: I see. I don't know about KMM module implementation (It required `id` only ).
I think we should adjust KMM module API interface too.

So let's remove `kmmModel` and use `id` only implementation 🙏🏼

myihsan: I would like to do it in another PR tomorrow since the change is over this PR's scope.

If there is a need to discuss with Android developers or to share that I will do it, please submit an issue and assign it to me. 🙏

ry-itto: > If there is a need to discuss with Android developers or to share that I will do it, please submit an issue and assign it to me. 🙏

It's okey!
I think we don't have to submit new issue about that.
So please make Pull Request about that 🙏🏼

ry-itto (approved):
LGTM 💯

PR #520, opened 2021-06-30
Display like audio spectrum during DroidKaigi FM playback

### LINKED ISSUE #381: Display like audio spectrum during DroidKaigi FM playback
## Kind (Required)

- Improvement

 ## Overview (Required)

- You don't have to actually make it from the sound. Is it possible to display an animation on the image during playback?

 ## Links	

-

rmakiyama:
## Issue
- close #381 

## Overview (Required)
- Display like audio spectrum during DroidKaigi FM playback.
- This animation is inspired by google podcast app.
- If you have better idea, please tell me feedback 😄 

## Links
-

## Screenshot

https://user-images.githubusercontent.com/7961496/123964292-3afc7700-d9ee-11eb-971e-d306da25ac17.mp4

### DIFF
--- a/uicomponent-compose/feed/src/main/java/io/github/droidkaigi/feeder/feed/FeedItem.kt
+++ b/uicomponent-compose/feed/src/main/java/io/github/droidkaigi/feeder/feed/FeedItem.kt
@@ -66,6 +66,7 @@ fun FeedItem(
             favoriteAnim,
             speakers,
             audioControl,
+            audioSpectrum,
         ) = createRefs()
         if (showMediaLabel) {
             Text(
@@ -114,6 +115,15 @@ fun FeedItem(
                 }
                 .clickable {
                     if (feedItem is FeedItem.Podcast) onClickPlayPodcastButton(feedItem)
+                }
+        )
+        AudioSpectrumAnimation(
+            isPlayingPodcast = isPlayingPodcast,
+            isVisible = feedItem is FeedItem.Podcast,
+            modifier = Modifier
+                .constrainAs(audioSpectrum) {
+                    linkTo(image.top, image.bottom, bias = 1.0f)
+                    linkTo(image.start, image.end, bias = 1.0f)
                 }
         )
         Text(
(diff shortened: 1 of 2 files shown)


takahirom:
👀

takahirom:
The animation looks cool! 🆒

uicomponent-compose/feed/src/main/java/io/github/droidkaigi/feeder/feed/FeedItem.kt
```diff
@@ -116,6 +117,15 @@ fun FeedItem(
...
         )
+        AudioSpectrumAnimation(
+            isPlayingPodcast = isPlayingPodcast,
+            isVisible = feedItem is FeedItem.Podcast,
+            modifier = Modifier
+                .constrainAs(audioSpectrum) {
+                    linkTo(image.top, image.bottom, bias = 1.0f)
+                    linkTo(image.start, image.end, bias = 1.0f)
+                }
+        )
```
takahirom, line 128: How about changing visibility like this? If we have isVisible in the component, I think the component will be complicated. If you have any problems with ConstraintLayout, please let me know.

```kotliin
if (feedItem is FeedItem.Podcast) {
          AudioSpectrumAnimation(
            isPlayingPodcast = isPlayingPodcast,
            modifier = Modifier
                .constrainAs(audioSpectrum) {
                    linkTo(image.top, image.bottom, bias = 1.0f)
                    linkTo(image.start, image.end, bias = 1.0f)
                }
        )
}
```

takahirom (approved):
Please let me know if you want to merge because it is only a detailed point.

uicomponent-compose/feed/src/main/java/io/github/droidkaigi/feeder/feed/FeedItem.kt
rmakiyama: > If we have isVisible in the component, I think the component will be complicated.

Sounds good! 😄 I think so too, thank you! I also think this component does not need `isPlayingPodcast`, what did you think?

On a different note, I think the same applies to `AudioControlButton` component about `isVisibility`, what do you think?

takahirom: Sounds good! That looks better! 👍 
How about fixing only AudioSpectrumAnimation now?

rmakiyama: I'll fix it! 🙌🏻

PR #528, opened 2021-07-03
[iOS] Create Media Detail Screen

ostk0069:
## Issue
- close #503

## Overview (Required)
- create MediaDetailScreen
- create SearchResultScreen (with considering empty result)
- fix search clear black view when start searching

## Links
-

## Screenshot
Before | After | After Search Result Empty | After Search Result Fill
:--: | :--: | :--: | :--:
 |  |  |

PR #532, opened 2021-07-04
[iOS] Commit snapshots with CI

ry-itto:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
- Commit snapshots on pushed to main branch (PR merged)
  - Commit only `ios/**` changed.
- Added because there was a problem running in the local environment
  - https://github.com/DroidKaigi/conference-app-2021/pull/528#issuecomment-873368433

## Links
-

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/ios/Tests/AboutFeatureTests/AboutFeatureTests.swift
+++ b/ios/Tests/AboutFeatureTests/AboutFeatureTests.swift
@@ -3,6 +3,10 @@ import TestUtils
 import XCTest
 
 final class AboutFeatureTests: XCTestCase {
+    override func setUp() {
+        initSnapshotTesting()
+    }
+    
     func testAboutScreen() {
         assertPreviewScreenSnapshot(AboutScreen_Previews.self)
     }
--- a/.github/workflows/ios-commit-snapshots.yml
+++ b/.github/workflows/ios-commit-snapshots.yml
@@ -28,7 +28,7 @@ jobs:
             buildSrc/src/**/*.kt
             **/gradle.properties
             gradle/**
-      - name Generate Strings & Colors & Assets
+      - name: Generate Strings & Colors & Assets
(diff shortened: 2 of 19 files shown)


ry-itto:
!github take-ios-snapshots

ry-itto:
!github take-ios-snapshots

ry-itto:
!github take-ios-snapshots

ios/Tests/AboutFeatureTests/AboutFeatureTests.swift
```diff
@@ -3,6 +3,10 @@ import TestUtils
 import XCTest
 
 final class AboutFeatureTests: XCTestCase {
+    override func setUp() {
```
ostk0069, line 6: It seems better to create custom XCTestCase class to make sure we don't have to add this everytime

ostk0069:
Please update the current README make sure contributors won't update the snapshots by themselves.

ios/Tests/AboutFeatureTests/AboutFeatureTests.swift
ry-itto: I don't think either will change.
When using custom XCTestCase, it is necessary to inherit so as not to forget.

ry-itto:
@ostk0069 
I fixed 🙏🏼 

> Maybe adding document comment in PreviewSnapshot.swift as well is way better
I think document comment in implementation file should be about implementation only not included operational comments.

ostk0069:
> I think document comment in implementation file should be about implementation only not included operational comments.

What I meant was adding some comment what contributors could stop creating snapshot  themselves. It does not matter it is as a document comment or just a simple comment.

ostk0069 (approved):

PR #526, opened 2021-07-02
[iOS] Home feature impl

ry-itto:
## Issue
- close None

## Overview (Required)
- [x] Add preview cases
- Fix FeedRepository to show contents

## Links
-

## Screenshot
Before | After
:--: | :--:
 |

PR #543, opened 2021-07-08
[iOS] Media Feature Impl 

ostk0069:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
- make sure repository connected to media screen

## Links
- related PR: https://github.com/DroidKaigi/conference-app-2021/pull/526

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/ios/Package.swift
+++ b/ios/Package.swift
@@ -122,6 +122,7 @@ var package = Package(
             dependencies: [
                 .target(name: "Component"),
                 .target(name: "Model"),
+                .target(name: "Repository"),
                 .target(name: "Styleguide"),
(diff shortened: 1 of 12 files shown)


ry-itto (approved):
There is a layout off like below (spacing). But it's okey. if it is not good, fix it with another issue.

スクリーンショット 2021-07-09 1 17 04

ostk0069:
> There is a layout off like below (spacing). But it's okey. if it is not good, fix it with another issue.

This problem seems to happen by using `GeometryReader` in `SmallCard`.
As you know, we are using `GeometryReader` to resize the image for `NukeUI.LazyImage` 
I will take a look about this with other opportunity.

PR #546, opened 2021-07-10
Add iOS snapshot files to Git LFS

### LINKED ISSUE #542: About iOS snapshot in main branch
## Kind (Required)

- Improvement

 ## Overview (Required)

- I was wondering if the method of committing a snapshot to the main branch in the repository is not so good because it image will remain as a git object forever. 👀 
- For example, you could create a `golden` branch and write automatic updates to it.
- I think it's okay to leave it as it is, but please tell us what you think. 🙇 

 ## Links	

-

takahirom:
## Issue
- close #542

## Overview (Required)
- Trying to store iOS snapshot to Git LFS
- I haven't rewritten past commits, so it won't disappear from the Git object, but future changes will no longer be saved. 👀 

## Links
-

## Screenshot
Before | After
:--: | :--:
 |


jmatsu:
@takahirom 
I also think that LFS can resolve the concern and it seems you have properly configured LFS for the snapshot files. However, this PR didn't trigger the snapshot testing, right? We should address the two points below.

- Make sure the workflow works with Git-LFS. We also enable LFS option of `actions/checkout` and add a step that pulls LFS files (and does cache, ideally).
- How to manage them should be documented. For example, LFS files won't be downloaded by default but they're required for iOS snapshot testing.

takahirom:
It seems that we should add `lfs: true` for `actions/checkout ` 👍 

We can add the document here.
https://github.com/DroidKaigi/conference-app-2021/tree/main/ios#snapshot-testing

.github/workflows/ios-test.yml
```diff
@@ -10,6 +10,8 @@ jobs:
     runs-on: macos-latest
     steps:
       - uses: actions/checkout@v2
+        with:
+          lfs: true
```
ry-itto, line 14: In iOS build job, screenshot files are not needed. So it's okey to remove this 👍🏼

.github/workflows/ios-test.yml
```diff
@@ -66,6 +70,8 @@ jobs:
...
```
ry-itto, line 74: ditto

> In iOS build job, screenshot files are not needed. So it's okey to remove this 👍🏼

.github/workflows/ios-test.yml
```diff
@@ -94,6 +100,8 @@ jobs:
...
```
ry-itto, line 104: ditto

> In iOS build job, screenshot files are not needed. So it's okey to remove this 👍🏼

ry-itto (commented):
Thank you for applying this with git-lfs 🙏🏼  
and currently, CI job for snapshot testing is failed.

I think applying commented change may work. Please try it 🙏🏼 

---

I have a concern about using `git-lfs`.
Using `git-lfs` can cost DroidKaigi organization money, is that okay?
( `git-lfs` bandwidth limit is 1GB/month for free. )

Reference:
https://docs.github.com/ja/billing/managing-billing-for-git-large-file-storage/about-billing-for-git-large-file-storage#

takahirom:
> I have a concern about using git-lfs.
Using git-lfs can cost DroidKaigi organization money, is that okay?
( git-lfs bandwidth limit is 1GB/month for free. )

Thank you for checking 🙏 
I think it is ok. 👍  But I don't want to waste it, so I would like to save it by reducing the image size. 👀

.github/workflows/ios-test.yml
takahirom:
```suggestion
```

.github/workflows/ios-test.yml
takahirom:
```suggestion
```

.github/workflows/ios-test.yml
takahirom:
```suggestion
```

takahirom:
Probably we can cache by using this 👀 
https://github.com/actions/checkout/issues/165#issuecomment-657673315

ry-itto (approved):
Looks good 👍🏼 

>  But I don't want to waste it, so I would like to save it by reducing the image size. 👀

If above is okey with caching `git-lfs` steps, please merge this 🙏🏼

takahirom:
It may be necessary to resize the image, so I would like to ask if that happens. 🙏

PR #551, opened 2021-07-11
Use LFS for snapshot update actions

takahirom:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
-Probably we should use lfs for snapshot update actions
- https://github.com/DroidKaigi/conference-app-2021/pull/550

## Links
-

## Screenshot
Before | After
:--: | :--:
 |


ry-itto (approved):
I missed 🙇🏼 
LGTM 👍🏼

PR #557, opened 2021-07-12
[iOS] Add `lfs: true` to checkout option

ry-itto:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
- Try to fix creating pull-request action.
- I think this example  https://github.com/actions/checkout/issues/165#issuecomment-657673315 is just a example. Maybe `lfs` option is needed.
  - But I don't know why `git lfs pull` works 😢 

## Links
-

## Screenshot
Before | After
:--: | :--:
 |


takahirom:
OK! Let's do it  👍

PR #564, opened 2021-07-22
[iOS] Remove the workaround for background colors

myihsan:
## Issue

It was an issue that breaks the navigation bar style when setting background by `background(_:alignment:)` with `ignoresSafeArea(_:edges:)`, so I worked around it by assigning background colors to `viewController.view`.
But, I can't reproduce the issue, and this workaround is not working in static preview.

If someone can reproduce the issue, please tell me.
I will try to fix it with a `ZStack`, which may be a better workaround @auramagi told me.

## Overview (Required)

- Remove the workaround for background colors (The update of snapshot test reference images should be done in other auto created PR?)

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/ios/Sources/MediaFeature/MediaScreen.swift
+++ b/ios/Sources/MediaFeature/MediaScreen.swift
@@ -29,6 +29,7 @@ public struct MediaScreen: View {
         NavigationView {
             WithViewStore(store) { viewStore in
                 ZStack {
+                    AssetColor.Background.primary.color.ignoresSafeArea()
                     ScrollView {
                         if viewStore.hasBlogs {
                             MediaSectionView(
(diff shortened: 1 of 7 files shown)


ostk0069:
Code looks great 😄 
Please fix the conflicts.

> The update of snapshot test reference images should be done in other auto created PR?

We still have problem with this. In this PR, it's good to go through.

myihsan:
Thanks for the title renaming!
I resolved the conflict.

ostk0069:
@myihsan 
CI failed.
We currently have problem with compiler. This could help you to fix https://github.com/DroidKaigi/conference-app-2021/blob/de107c284561b8a50071675d5d25d4fd6276738c/ios/Sources/HomeFeature/HomeScreen.swift#L36

ios/Sources/MediaFeature/MediaScreen.swift
```diff
@@ -29,6 +29,7 @@ public struct MediaScreen: View {
         NavigationView {
             WithViewStore(store) { viewStore in
                 ZStack {
+                    AssetColor.Background.primary.color.ignoresSafeArea()
```
myihsan, line 32: This is actually broken by the animation of the `ZStack`.
But I found out that the animation can cause other problems, so I remove it in #571.

ostk0069 (approved):

PR #571, opened 2021-07-25
[iOS] Use view states

myihsan:
## Issue

The `body` of almost every `View` is executed every time the `AppState` changes. So there are a lot of unnecessary executions (thanks to SwiftUI, it doesn't cost much to redraw the View, but the `body` is executed all over the place).

I don't have enough time to describe this issue in detail, so, please ask me which part is not clear directly🙇‍♂️

## Overview (Required)

- Avoid unnecessary `body` execution by `scope(state:action:)` (only accomplished in `MediaScreen` because of the limited time)
- Avoid using animation(_:)
- Avoid a weird animation when `isSearchResultVisible` and `searchedFeedContents.isEmpty` change in the same time

## Links

- https://www.pointfree.co/episodes/ep151-composable-architecture-performance-view-stores-and-scoping

### DIFF
--- a/ios/Sources/Component/View/FeedContentListView.swift
+++ b/ios/Sources/Component/View/FeedContentListView.swift
@@ -38,7 +38,6 @@ public struct FeedContentListView: View {
             }
         }
         .padding(.horizontal, SmallCard.Const.margin)
-        .animation(.easeInOut)
     }
 }
 
--- a/ios/Sources/MediaFeature/MediaFeature.swift
+++ b/ios/Sources/MediaFeature/MediaFeature.swift
@@ -7,21 +7,18 @@ public struct MediaState: Equatable {
     // In order not to use any networks for searching feature,
     // `feedContents` is storage to search from & `searchedFeedContents` is searched result from `feedContents`
     public var feedContents: [FeedContent]
-    public var searchedFeedContents: [FeedContent]
-    var isSearchResultVisible: Bool
+    public var searchedFeedContents: [FeedContent]?
     var isSearchTextEditing: Bool
     var moreActiveType: MediaType?
 
     public init(
         feedContents: [FeedContent],
-        searchedFeedContents: [FeedContent] = [],
-        isSearchResultVisible: Bool = false,
+        searchedFeedContents: [FeedContent]? = nil,
         isSearchTextEditing: Bool = false,
         moreActiveType: MediaType? = nil
     ) {
         self.feedContents = feedContents
         self.searchedFeedContents = searchedFeedContents
-        self.isSearchResultVisible = isSearchResultVisible
         self.isSearchTextEditing = isSearchTextEditing
         self.moreActiveType = moreActiveType
     }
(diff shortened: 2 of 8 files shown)


ios/Sources/FavoritesFeature/FavoritesScreen.swift
```diff
@@ -15,20 +15,23 @@ public struct FavoritesScreen: View {
...
-                            },
-                            tapFavorite: { isFavorited, contentId in
-                                viewStore.send(.tapFavorite(isFavorited: isFavorited, id: contentId))
-                            }
-                        )
+            ZStack {
+                AssetColor.Background.primary.color.ignoresSafeArea()
+                WithViewStore(store) { viewStore in
+                    if viewStore.feedContents.isEmpty {
+                        Text("表示するコンテンツがありません")
```
ostk0069, line 22: In the latest `main` branch this is not used any more. We currently use [`FavoritesEmptyView`](https://github.com/DroidKaigi/conference-app-2021/blob/f31566b6c84ddbfe4f40f530f2a9f6ffeef3fbb6/ios/Sources/FavoritesFeature/FavoritesScreen.swift#L22)
I thought this should be conflicted but it's not.
Could you please fix to follow the latest `main` branch?

ios/Sources/Component/View/FeedContentListView.swift
```diff
@@ -38,7 +38,6 @@ public struct FeedContentListView: View {
             }
         }
         .padding(.horizontal, SmallCard.Const.margin)
-        .animation(.easeInOut)
```
ostk0069, line 41: First time show mediaScreen, it has rattling with search bar and feedlist.
Do you know anything about this?

ios/Sources/MediaFeature/MediaFeature.swift
```diff
@@ -7,26 +7,24 @@ public struct MediaState: Equatable {
     // In order not to use any networks for searching feature,
     // `feedContents` is storage to search from & `searchedFeedContents` is searched result from `feedContents`
     public var feedContents: [FeedContent]
-    public var searchedFeedContents: [FeedContent]
-    var isSearchResultVisible: Bool
+    public var searchedFeedContents: [FeedContent]?
```
ostk0069, line 10: I want to know why this should be optional 🙇

ios/Sources/FavoritesFeature/FavoritesScreen.swift
myihsan: It's weird to me that there is no conflict here.
I will rebase soon.

ios/Sources/Component/View/FeedContentListView.swift
myihsan: I notice that, but it seems like an issue of iOS when combining SwiftUI with UIKit.
This line can improve the rattling but also introduce other issues:
- Unnecessary animation when scrolling to make the navigation bar become the inline style
- Break the navigation appearance when setting the background color by SwiftUI (may be solved by embedding them in an extra `ZStack`)

So I think may be better to ignore the rattling.

ios/Sources/MediaFeature/MediaFeature.swift
myihsan: To avoid the condition below.
```swift
searchText.isEmpty && searchedFeedContents.isEmpty && isSearchResultVisible
```
This condition seems impossible but it actually exists during the animation.

You may notice the issue by revert this change and run the app on a simulator that enabled slow animations.

ostk0069 (approved):
LGTM!

PR #585, opened 2021-08-01
[iOS] Reduce snapshot image size

ry-itto:
## Issue
- close None

## Overview (Required)
- Reduce iOS Snapshot image size
  - For example, 166KB -> 16KB (in my environment)
- Fix test fails

## Links
-

## Screenshot
Before | After
:--: | :--:
 |


takahirom:
> Snapshot does not match reference

Probably it is OK, isn't it?

ry-itto:
> GitHub Actions:
> 
> 
> 
> > Snapshot does not match reference
> 
> 
> 
> Probably it is OK, isn't it?

Its okey!
Update current snapshots is needed to pass snapshot testing CI.

takahirom (approved):
LGTM!

PR #593, opened 2021-08-24
Performance improvement for feed list

takahirom:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)

* Add key property for LazyColumn
* Do not pass property(listState) that are updated frequently
Instead, use derivedStateOf

## Links
-

## Screenshot
Before | After
:--: | :--:
 |


NUmeroAndDev:
@takahirom 
For this code, I think it is better not to surround with `item{}` with `if ()` because the `listState` will be changed and recomposed by increasing or decreasing the item. 🤔 

```
item(key = lazyColumnRobotKey) {
  if (isLastItemVisible) {
      ...
  }
}
```

takahirom:
I will check it! Thank you for your information 🙏

takahirom:
I don't know if I should look at this, it looks like the recompose is decreasing when scrolling. 👀 
image

PR #600, opened 2021-09-02
Settings exported attribute

NUmeroAndDev:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
- If this app targets Android 12, you need to explicitly declare the `android:exported` attribute.
- This app is targets Android 11, but I think no problem to support it in advance.

## Links
- https://developer.android.com/about/versions/12/behavior-changes-12#exported

## Screenshot
Before | After
:--: | :--:
 |


takahirom:
👀

takahirom:
LGTM!! Nice tips!

Issue #628, opened 2021-09-20
Database cache of timetable item


takahirom:
## Kind (Required)

- Improvement

 ## Overview (Required)

- We will put the timetable items in the Database and use it to display the data retrieved from the API.
- Basically, we will change the data from the Database data to flow through Coroutiens Flow and use it in the application.

https://github.com/DroidKaigi/conference-app-2021/blob/2bbcbed97b2c7a49540e0519ed2ccbbdcd235df3/data/repository/src/commonMain/kotlin/io/github/droidkaigi/feeder/data/TimetableRepositoryImpl.kt#L10

 ## Links	

-

koji-1009:
✋

takahirom:
Thanks🙏🙏Assigned!!

koji-1009:
Hi, @takahirom san.

I have a question. Could you discuss a little?

---

The code sets `io.github.droidkaigi.timetableer.data`, but the actual file path is `io.github.droidkaigi.feeder.data`.
From what I've read in the code, it certainly looks like a different file path would be a good idea. However, this will also affect the repository module and other issue(#624). 🤔 

Do you have any plan about this?

https://github.com/DroidKaigi/conference-app-2021/blob/2bbcbed97b2c7a49540e0519ed2ccbbdcd235df3/data/repository/src/commonMain/kotlin/io/github/droidkaigi/feeder/data/TimetableRepositoryImpl.kt#L1

takahirom:
👀

koji-1009:
Thanks!

PR #670, opened 2021-09-25
Changing variable names

koji-1009:
## Issue
- ref #628

## Overview (Required)
- I've quoted the code directly from [the reference](https://github.com/DroidKaigi/conference-app-2021/blob/b7c54f903562d9d0cf8ada179c20200af1ea6dec/data/db/src/commonMain/kotlin/io/github/droidkaigi/feeder/data/FeedItemDao.kt#L202) and it's not in context, so I'll fix it. 🙏 

## Links
-

## Screenshot
Before | After
:--: | :--:
 |


takahirom (approved):
LGTM📛

PR #683, opened 2021-09-27
[iOS] Fix image tap enable

ostk0069:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
- when tap image of content which should open webView, but it does not. so fix it up.

## Links
-

## Screenshot
Before | After
:--: | :--:
 |


ry-itto (approved):
Thanks 👍🏼

PR #687, opened 2021-09-28
Use official way of creating xcframework 

ry-itto:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
- From Kotlin 1.5.30, creating xcframework is officially supported.(https://kotlinlang.org/docs/whatsnew1530.html#support-for-xcframeworks)
  - So I apply that way.
- Default output directory is `project.buildDir` and it cannot be changed (I think). So I add create xcframework with copy it to `ios/build` directory task.
- After this, I will try Apple Silicon(ios-simulator-arm64) support 🙏🏼 

## Links
-

## Screenshot
Before | After
:--: | :--:
 |


takahirom (approved):
Thanks as always🙏

ry-itto:
🙏🏼

Issue #692, opened 2021-09-29
Add insets padding to TimetableItem list


takahirom:
**Overview (Required)**
inset padding to timetable item list so the item is cropped

**Steps To Reproduce**

1. Go to 'TimetableScreen'
2. Scroll
4. See cropped item

**Expected behavior**
not cropped

**Screenshots**

![image](https://user-images.githubusercontent.com/1386930/135245141-24dd822a-80e2-4d3e-8424-dddce4808a8f.png)

**Env:**
 - Device: [e.g. Pixel3]
 - OS: [e.g. 9.0]
 - App Version or the branch [e.g. 0.1.0 or master]

**Additional context**

You can use insetter for this issue.
https://github.com/DroidKaigi/conference-app-2021/blob/main/uicomponent-compose/feed/src/main/java/io/github/droidkaigi/feeder/feed/FeedScreen.kt#L309

HironobuIga:
I'd like to challenge :)

takahirom:
Thanks! Assigned 👍

PR #711, opened 2021-10-02
Add system bar padding to TimetableDetailScreen

kseito:
## Issue
-

## Overview (Required)
- I found an issue that the bottom of the TimetableDetailScreen is cropped.
- I think the cause of this issue is a lack of consideration of padding.
- So I added system bottom bar padding to TimetableDetailScreen.
- This is similar to #692 .

### Steps To Reproduce
1. Go to 'TimetableDetailScreen'.
2. Scroll to the bottom of the screen.
3. See cropped item.

## Links
-

## Screenshot
Before | After
:--: | :--:
 |


takahirom (approved):
LGTM 👍 Thank you for catching the issue!!

PR #721, opened 2021-10-04
Fix language setting apply refactor

ry-itto:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
- Change language setting from Setting App
- I gave up to change language setting in App.
  - When App calls KMM Repository module method, a SettingScreen's sheet is unintentionally closed. I don't know that reason.
  - When I adopt language setting to whole App, API responses looks good, but `Localizable.strings` labels are not good. (Not changed.)
    - In case we want to apply `Localizable.strings` labels to App forcibly, it is needed to restart App, however I don't know how to do it (When I searched on Google, I obtained a result that is impossible).
- I'm sorry for getting this result @yuukiw00w 
  - Kotlin side implementations are remained. I would appreciate it if you could look forward to its implementation.

## Links
-

## Screenshot
Before | After
:--: | :--:
 |


ios/Sources/Repository/LanguageRepository.swift
```diff
@@ -19,11 +19,13 @@ public struct LanguageRepository: LanguageRepositoryProtocol, KMMRepositoryProto
...
-            scopeProvider: scopeProvider
-        )
-        .map { _ in }
+        Future { promise in
+            repository.changeLanguage(language: language.kmmLang).subscribe(scope: scopeProvider.scope) { _ in
+                promise(.success(()))
+            } onFailure: { error in
+                promise(.failure(KotlinError.fetchFailed(error.description())))
+            }
+        }
```
ostk0069, line 28: do we need this diff?

ios/Sources/SettingFeature/SettingScreen.swift
```diff
@@ -17,61 +16,16 @@ public struct SettingScreen: View {
     }
 
     public var body: some View {
-        WithViewStore(store) { viewStore in
+        WithViewStore(store) { _ in
```
ostk0069, line 19: maybe we could remove whole screen?

ios/Sources/TimetableFeature/Detail/TimetableDetailScreen.swift
```diff
@@ -23,15 +23,17 @@ public struct TimetableDetailScreen: View {
         ZStack {
             ScrollView {
                 WithViewStore(store) { viewStore in
-                    Text(viewStore.timetable.title.get(by: viewStore.language))
+                    Text(viewStore.timetable.title.get(
+                            by: Foundation.Locale.current.language
```
ostk0069, line 27: does this really works when change language dynamically?

ios/DroidKaigi 2021/DroidKaigi 2021/en.lproj/Localizable.strings
```diff
@@ -0,0 +1,2 @@
+
```
ostk0069, line 1: Why do we need these files?

ry-itto: To show language setting in Setting App. 
If this files are not existed, no settings found in Setting App.

ios/Sources/TimetableFeature/Detail/TimetableDetailScreen.swift
ry-itto: Yes.

ios/Sources/SettingFeature/SettingScreen.swift
ry-itto: I want to remain this screen 🙏🏼 
Because there is a possibility about implementing change color theme feature.

ios/Sources/Repository/LanguageRepository.swift
ry-itto: This is no needed 🙏🏼 
I'll revert this 👍🏼

ios/DroidKaigi 2021/DroidKaigi 2021/en.lproj/Localizable.strings
ostk0069: Because we have done the language from setting, color theme could be the same I guess.
If we have time to implement this, I think it's okay to keep like this.

ios/Sources/SettingFeature/SettingScreen.swift
ostk0069: Because we have done the language from setting, color theme could be the same I guess.
If we have time to implement this, I think it's okay to keep like this.

ostk0069 (approved):

